PR #14, opened 2023-01-01
updates in decisions

markus2330:
@buenaflor can you have a look at the updated alternatives? Is there something new for you which you find worth considering?

What do you think about Rust Diesel?

I also did a few other clarifications (stable and not nightly Rust).

### DIFF
--- a/doc/decisions/2d_drawing_lib.md
+++ b/doc/decisions/2d_drawing_lib.md
@@ -2,7 +2,7 @@
 
 ## Problem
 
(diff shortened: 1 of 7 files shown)


buenaflor:
> Rust Diesel

I did some little research and Diesel looks good to me.

> Updated alternatives, worth considering

I am currently rewriting the frontend language decision to be in TypeScript instead of JavaScript. It makes maintainability, refactoring, code reviews etc... much easier. Since it is a superset of JavaScript we still have access to all the features like before.

> Tamagui

This looks interesting, I'll take a look at it

PR #17, opened 2023-01-05
more usecases

buenaflor:
Contains usecases for:

-  entry+list of which seeds the user has (for better selection of plants), including amount and expire dates (higher recommendation of using the seed if it expires soon)
-  searching of maps via keywords (including public maps of other users)
-  creation/deletion of map (deletion with undo for a month)
-  timeline: scroll bar (granularity month) and date field to change the time that is currently visible on the map
-  label layer

### DIFF
--- a/doc/usecases/entry_list_seeds.md
+++ b/doc/usecases/entry_list_seeds.md
@@ -5,7 +5,7 @@
 - **Scope:** Seed Management
 - **Level:** User Goal
 - **Actors:** App User
-- **Brief:** The user can manage their seeds by adding, viewing, and deleting them in the app.
+- **Brief:** The users can manage their seeds by adding, viewing, and deleting them in the app.
 - **Status:** Draft
 
 ## Scenarios
@@ -15,12 +15,11 @@
 - **Main success scenario:**
   The user is able to add new seeds to their list by specifying the name, amount, and expiration date. 
   They can also view and delete seeds from their list. 
-  The app will also likely recommend using seeds that are about to expire.
 - **Alternative scenario:**
-  The user accidentally deletes a seed from their list and uses the seed entry to reinsert it.
+  The user accidentally deletes a seed from their list and uses the seed insert feature to correct it.
 - **Error scenario:**
   There is an error when the user attempts to add a seed to their list, such as an invalid expiration date or a seed with the same name already existing. 
-  In this case, the app displays an error message to the user.
+  In this case, the app validates while the user is typing and shows a validation error.
 - **Postcondition:**
   The user's seed list is updated with the added, deleted, or restored seeds.
 - **Non-functional Constraints:**
--- a/doc/usecases/entry_list_seeds.md
+++ b/doc/usecases/entry_list_seeds.md
@@ -1,0 +1,26 @@
+# Use Case: Entry and list of seeds
+
+## Summary
+
+- **Scope:** Seed Management
+- **Level:** User Goal
+- **Actors:** App User
+- **Brief:** The user can manage their seeds by adding, viewing, and deleting them in the app.
+- **Status:** Draft
+
+## Scenarios
+
+- **Precondition:**
+  The user has opened the seed management feature in the app.
+- **Main success scenario:**
+  The user is able to add new seeds to their list by specifying the name, amount, and expiration date. 
+  They can also view and delete seeds from their list. 
+  The app will also likely recommend using seeds that are about to expire.
+- **Alternative scenario:**
+  The user accidentally deletes a seed from their list and uses the seed entry to reinsert it.
+- **Error scenario:**
+  There is an error when the user attempts to add a seed to their list, such as an invalid expiration date or a seed with the same name already existing. 
+  In this case, the app displays an error message to the user.
+- **Postcondition:**
+  The user's seed list is updated with the added, deleted, or restored seeds.
+- **Non-functional Constraints:**
--- a/doc/usecases/label_layer.md
+++ b/doc/usecases/label_layer.md
@@ -18,7 +18,7 @@
 - **Alternative scenario:**
   The user accidentally adds a label in the wrong location and uses the app's undo function to correct the mistake
 - **Error scenario:**
-  The user tries to add a label with more than 2000 characters, and the app displays an error message indicating the maximum character limit.
+  The user tries to add a label with more than 2000 characters and the app prevents it by stopping any further insertions into the label field.
 - **Postcondition:**
   The user's map includes labels as desired.
 - **Non-functional Constraints:**
--- a/doc/usecases/label_layer.md
+++ b/doc/usecases/label_layer.md
@@ -20,5 +20,5 @@
 - **Error scenario:**
   The user tries to add a label with more characters than the maximum allowed, and the app displays an error message indicating the maximum character limit.
 - **Postcondition:**
-  The user's landscape plan includes labels as desired.
+  The user's map includes labels as desired.
 - **Non-functional Constraints:**
--- a/doc/usecases/label_layer.md
+++ b/doc/usecases/label_layer.md
@@ -1,0 +1,24 @@
+# Use Case: Label Layer
+
+## Summary
+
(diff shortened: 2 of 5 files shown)


doc/usecases/entry_list_seeds.md
```diff
@@ -0,0 +1,26 @@
+# Use Case: Entry and list of seeds
+
+## Summary
+
+- **Scope:** Seed Management
+- **Level:** User Goal
+- **Actors:** App User
+- **Brief:** The user can manage their seeds by adding, viewing, and deleting them in the app.
```
markus2330, line 8:
```suggestion
- **Brief:** The users can manage their seeds by adding, viewing, and deleting them in the app.
```

doc/usecases/entry_list_seeds.md
```diff
@@ -0,0 +1,26 @@
...
+- **Status:** Draft
+
+## Scenarios
+
+- **Precondition:**
+  The user has opened the seed management feature in the app.
+- **Main success scenario:**
+  The user is able to add new seeds to their list by specifying the name, amount, and expiration date. 
+  They can also view and delete seeds from their list. 
+  The app will also likely recommend using seeds that are about to expire.
```
markus2330, line 18: This is to be added in the Planting use case, not here.

doc/usecases/entry_list_seeds.md
```diff
@@ -0,0 +1,26 @@
...
+- **Alternative scenario:**
+  The user accidentally deletes a seed from their list and uses the seed entry to reinsert it.
```
markus2330, line 20: Please clarify "seed entry".

doc/usecases/entry_list_seeds.md
```diff
@@ -0,0 +1,26 @@
...
+- **Error scenario:**
+  There is an error when the user attempts to add a seed to their list, such as an invalid expiration date or a seed with the same name already existing. 
+  In this case, the app displays an error message to the user.
```
markus2330, line 23: Ideally we also have validation while typing, directly in the text fields.

doc/usecases/label_layer.md
```diff
@@ -0,0 +1,24 @@
+# Use Case: Label Layer
```
markus2330, line 1:
```suggestion
# Use Case: Labels Layer
```

doc/usecases/label_layer.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Summary
+
+- **Scope:** Label Layer
```
markus2330, line 5:
```suggestion
- **Scope:** Labels Layer
```

doc/usecases/label_layer.md
```diff
@@ -0,0 +1,24 @@
...
+- **Level:** User Goal
+- **Actors:** App User
+- **Brief:** The user can add labels to their map to identify different elements or areas
```
markus2330, line 8:
```suggestion
- **Brief:** The user adds labels to their map to identify different elements or areas
```

doc/usecases/label_layer.md
```diff
@@ -0,0 +1,24 @@
...
+- **Status:** Draft
+
+## Scenarios
+
+- **Precondition:**
+  The user has opened the app and selected the label layer.
```
markus2330, line 14:
```suggestion
  The user has opened the app and selected the labels layer.
```

doc/usecases/label_layer.md
```diff
@@ -0,0 +1,24 @@
...
+- **Main success scenario:**
+  The user successfully adds labels to their map by selecting the element or area they want to label and entering the desired text in the label field.
```
markus2330, line 16:
```suggestion
  The users successfully add labels to their map by selecting the element or area they want to label and entering the desired text in the label field.
```

doc/usecases/label_layer.md
```diff
@@ -0,0 +1,24 @@
...
+  The label is displayed on the map and can be edited or removed at any time.
+- **Alternative scenario:**
+  The user accidentally adds a label in the wrong location and uses the app's undo function to correct the mistake
+- **Error scenario:**
+  The user tries to add a label with more characters than the maximum allowed, and the app displays an error message indicating the maximum character limit.
```
markus2330, line 21: Is the limit needed? How long will that limit be.

doc/usecases/map_search.md
```diff
@@ -0,0 +1,27 @@
...
+- **Brief:** The user can search for maps within the app using keywords, including the ability to search for public maps created by other users.
+- **Status:** Draft
+
+## Scenarios
+
+- **Precondition:**
+  The user has opened the app and is on the map search page.
+- **Main success scenario:**
+  The user enters a keyword into the search field and the app displays a list of maps that match the keyword, including both the user's own maps and public maps created by other users. 
+  The user can then select a map to view it in more detail.
```
markus2330, line 17:
```suggestion
  The user can then select a map to view it.
```

doc/usecases/map_timeline.md
```diff
@@ -0,0 +1,26 @@
...
+- **Actors:** App User
+- **Brief:** The user can view the map at different points in time by using the timeline feature
+- **Status:** Draft
+
+## Scenarios
+
+- **Precondition:**
+  The user has opened the app and selected the map view.
+- **Main success scenario:**
+  The user selects the timeline view and uses the scroll bar with a month granularity or date field to navigate to a different point in time.
```
markus2330, line 16: Is it possible that the scroll bar visually indicates in which months there are changes and at which months the map is empty (without plants)?

doc/usecases/map_timeline.md
```diff
@@ -0,0 +1,26 @@
...
+## Scenarios
+
+- **Precondition:**
+  The user has opened the app and selected the map view.
+- **Main success scenario:**
+  The user selects the timeline view and uses the scroll bar with a month granularity or date field to navigate to a different point in time. 
+  The map updates to show the state of the garden at the selected time.
+- **Alternative scenario:**
+  The user tries to navigate to a date that is outside the range of the timeline. 
+  In this case, the app displays an error message indicating that the requested date is not available.
```
markus2330, line 20: What is the range? For reasonable dates it must show empty map, the user might want to draw there.

markus2330 (changes requested):
Great work! All use cases created within the milestone! :1st_place_medal:

doc/usecases/label_layer.md
buenaflor: I think a sensible limit is not bad since we can also prevent any possible attacks by users trying to overload the system with an infinite amount of strings.

My first thought is to use around 2,000 characters, that should be between 285 words and 500 words and I think that's plenty for labeling

doc/usecases/map_timeline.md
buenaflor: Should be possible yes

doc/usecases/map_timeline.md
buenaflor: Hm, I'm not sure here. Do you have a suggestion on how to handle this?

doc/usecases/label_layer.md
markus2330: Doesn't some token or similar (that only logged in users get) already protect against this? But as the labels are directly shown on the map, I also don't see any possibility that 2000 characters per label wouldn't be enough.

Nevertheless, I would prefer if the UI simply prevents to enter more than 2000 characters and the error message (needed on API level) actually should never be seen by the user.

doc/usecases/map_timeline.md
markus2330: That would be amazing :heart:

doc/usecases/map_timeline.md
markus2330: Difficult to judge without trying it out. Yvonne was thinking about only showing 1,5 years in total in the timeline, I was thinking about all events that have ever happened (can be e.g. 10 years) and 1 year additionally at beginning and end.

What definitely would be nice is to get mouseover hints which month would be selected on a click.

PR #29, opened 2023-01-17
use cases v3

markus2330:
Here one layer is removed (animals), so we are again at the same number of layers. :wink: 

This is probably still not the final version. But better to find all problems now.


README.md
```diff
@@ -9,51 +9,50 @@ PermaplanT is an app for
...
+  - what is seasonal
+  - increasing diversity
 
 ## Main Features
 
 - users can have any number of maps
 - users can work on the same map together
 - maps have a fixed number of layers (see below)
-- undo/redo
+- undo/redo within current session for changes in the map
```
buenaflor, line 31: Is this managed locally in the frontend? I assume sessions start when a user visits the website and ends when the user leaves the website

buenaflor (approved):

README.md
markus2330: I wrote it because I thought it is easier for you to implement. It is also like how I understood what you originally wrote in undo_redo usecase.

From user perspective best would be to be able to undo/redo everything in the last month. But this makes the database quite complicated. But if you know an easy solution, Yvonne would be happy.

PR #32, opened 2023-01-22
Use cases v4

markus2330:


markus2330:
Cookies is not a must, let us see if the static URL idea even works. If you know "no cookies" and/or static URL doesn't work, please say so. Otherwise we will talk about it in the next meeting anyway.

markus2330:
Should be okay now.

aidnurs (approved):

Issue #47, opened 2023-02-09
fetch German common name from wikipedia


markus2330:
If I enter, e.g., https://de.wikipedia.org/wiki/Acanthus_mollis I get the German name "Wahrer Bärenklau". @aidnurs can we easily use this to pre-fill the German common name column?

markus2330:
Have you already looked into this one? This would make Yvonne's work of adding additional data much more comfortable.

markus2330:
scraper/data/detail.csv still doesn't seem to contain German names?

When I rerun the scraper, I don't get the same results, e.g. first row:

```diff
 "Folder Name","Common Name","Common Name DE","Subfamily","To check","Is Tree","Mature Size Height","Mature Size Width","Nutrition Demand","Article Last Modified At","Is Variety","Binomial name","Genus","Family","Edible uses","Medicinal uses","Material uses &amp; Functions","Botanic","Propagation","Cultivation","Environment","Material uses","Functions","Provides forage for","Provides shelter for","Hardiness Zone","Heat Zone","Water","Sun","Shade","Soil PH","Soil Texture","Soil Water Retention","Environmental Tolerances","Native Climate Zones","Adapted Climate Zones","Native Geographical Range","Native Environment","Ecosystem Niche","Root Zone Tendancy","Deciduous or Evergreen","Herbaceous or Woody","Life Cycle","Growth Rate","Fertility","Pollinators","Flower Colour","Flower Type","Plant References","Has drought tolerance","Tolerates wind"
-"Allium_cepa_ascalonicum","Shallot",,"","Is Variety",,"0.3",,,"2013-05-04T10:07:00.000Z",true,"Allium cepa ascalonicum","Allium","Alliaceae",,,"
-[6][7][8][9]
-","[18]","","","",,"","","","5",,"moderate","full sun","no shade","
-acid, neutral, alkaline, very alkaline
+"Aconitum_dienorrhizum","",,"",,,,,,"2013-05-04T10:02:00.000Z",,"Aconitum dienorrhizum","Aconitum","Ranunculaceae",,,"","[9]","","","",,"","","",,,"moderate","full sun","light shade","
+acid, neutral, alkaline
```

aidnurs:
Hey @markus2330,

[...]

PR #48, opened 2023-02-10
Define database layout for plant hierarchy

aidnurs:


scraper/data/detail_yvonne.csv
```diff
@@ -0,0 +1,31 @@
+Folder Name,Common Name,Common Name DE,Subfamily,To check,Is Tree,Mature Size Height,Mature Size Width,Nutrition Demand,Binomial name,Genus,Family,Edible uses,Medicinal uses,Material uses &amp; Functions,Botanic,Propagation,Cultivation,Environment,Material uses,Functions,Provides forage for,Provides shelter for,Hardiness Zone,Heat Zone,Water,Sun,Shade,Soil PH,Soil Texture,Soil Water Retention,Environmental Tolerances,Native Climate Zones,Adapted Climate Zones,Native Geographical Range,Native Environment,Ecosystem Niche,Root Zone Tendancy,Deciduous or Evergreen,Herbaceous or Woody,Life Cycle,Growth Rate,Fertility,Pollinators,Flower Colour,Flower Type,Has drought tolerance,Tolerates wind,Plant References
```
markus2330, line 1: Please add a README.md how this merging works (what is Yvonne allowed to change etc.)

markus2330 (changes requested):
Good work!

doc/decisions/database_structure.md
```diff
@@ -0,0 +1,78 @@
...
+
+The main goal of this document is to describe the database structure/layout especially in terms of relations between data entities.
+
+## Problem
+
+Semantically the plant information can be described as a hierarchy of entities. E.g., a plant is always a member on the variety level, while a variety is a part of species, species is a part of genus, genus is a part of subfamily, subfamily is a part of family. Please take a look at the diagram below. Each of these entities can have their own properties, e.g., a variety under a specie can have different height than the specie that it belongs to.
+
+The challenge is to define the structure of the database in a way that the data can be easily queried and updated.
+
+Another goal is to be able to define relations between entities, e.g., a plant of one variety doesn't want to be neighbor to plant of some other variety.
```
markus2330, line 11: Actually this is a Constraint for the decision, like all other features we want for the hierarchy.

doc/decisions/database_structure.md
```diff
@@ -0,0 +1,78 @@
...
+
+> Table inheritance is typically established when the child table is created, using the INHERITS clause of the CREATE TABLE statement.
+
+So the inheritance is useful to deal with complex DDL structure on the startup, but will not help us to avoid bulk operations e.g. updating a column for every `variety` in the entire `genus`
+
+## Decision
+
+1. Hierarchy
+
+Hierarchy of entities should be implemented using foreign keys and table relationships instead of Inheritance feature of PSQL. In case of a conflict, the lowest entity level in the hierarchy is prioritized. E.g. when there is a height value in a species and in a variety, variety wins.
```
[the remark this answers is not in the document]
markus2330: Actually it would be best if you can give a few SQL commands how we would update and select attributes from the plants.

markus2330 (changes requested):
Additional comments.

doc/decisions/database_structure.md
```diff
@@ -0,0 +1,78 @@
...
+
+Plant relations(e.g. like, dislike, etc.) should be implemented using separate many-to-many tables. This is because the relations are not part of the plant entity, but rather a property of the relationship between two plants.
+
+2. Metatables for parent entities i.e. genus, subfamily, family
```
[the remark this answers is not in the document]
markus2330: She agrees, we can always assume that a species can be planted, even though often not the species itself is relevant for the garden but its varieties instead. However, the species theoretically exists as a plant. We can solve this nicely in the UI by having a collapsing entry (clicking on "+" shows varieties).

The solution of https://docs.trefle.io/docs/guides/getting-started/ (having a botanically-non-existing hierarchy layer "Plant") is more confusing than what it might help.

doc/decisions/database_structure.md
```diff
@@ -0,0 +1,78 @@
...
+
+    class Family:::abstract {
+    }
+    class Subfamily:::abstract {
+    }
+    class Genus:::abstract {
+    }
+    class Species {
+      +int height
+      +int width
```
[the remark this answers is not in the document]
aidnurs: explicitly mentioned in the doc once again that the diagram provides an example of such attribute classification.

doc/decisions/database_structure.md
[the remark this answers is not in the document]
aidnurs: I agree with you that we should explicitly document table relationships. Since this document is a decision document I would like to keep technical details out of it i.e. the better place would be [database_schemata.md](https://github.com/ElektraInitiative/PermaplanT/blob/db-docs/doc/architecture/database_schemata.md) in my opinion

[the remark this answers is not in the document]
markus2330: If the details are part of what is to be decided, they need to be mentioned. (If the details are not relevant for the decision, they don't need to be mentioned.)

doc/decisions/database_structure.md
```diff
@@ -0,0 +1,95 @@
...
+- subfamily belongs to a family.
+
+Please take a look at the diagram below.
+
+Each of these entities can have their own properties.
+E.g., a variety under a specie can have different height than the specie that it belongs to.
+
+The challenge is to define the structure of the database in a way that the data
+- can be easily queried and
+. updated without redundancy (e.g. setting attributes for a whole family).
```
markus2330, line 26:
```suggestion
- updated without redundancy (e.g. setting attributes for a whole family).
```

doc/decisions/database_structure.md
```diff
@@ -0,0 +1,95 @@
...
+- on variety level, and/or
+- on species level.
+
+Furthermore:
+- a variety belongs to a specie
+- specie belongs to a genus
+- genus belongs to a subfamily or family
+- subfamily belongs to a family.
+
+Please take a look at the diagram below.
```
markus2330, line 19: Please directly add it here. (Forward references are bad style, especially if they are unnamed.)

markus2330 (changes requested):
Looked at decision again. It is probably better in future to have decisions in separate PRs, as they sometimes have heavy discussions.

aidnurs:
Hey @markus2330,

I was stuck and confused with this decision for a bit because I couldn't say for sure what the foreign key relationships will look like since I don't have enough data and insights for now. It is going to depend on the semantics of the data in the future. Let me provide you an example here: 

all the information scraped from practicalplants is represented in a flat form e.g. there is no semantical difference between entries yet(species vs genus vs....) i.e. we should consider them all as varieties for now. I was able to extract all unique occurrences of genus and family values, however, the species list is still empty. So the relationship structure cannot be fixed yet and will take some time. 

I totally agree with you in your latest comment that we should keep decisions apart from technical parts, I was quite confused with the decision structure in general since I thought that a general decision would be sufficient e.g. PSQL Inheritance was rejected right away and flat table structure was preferred with the first commit. I suggest we proceed with the current state and do not fix the structure yet, so it doesn't hold our progress. 

Let me know what you think. Thanks!
Nursultan

doc/decisions/database_structure.md
```diff
@@ -0,0 +1,78 @@
...
+So the inheritance is useful to deal with complex DDL structure on the startup, but will not help us to avoid bulk operations e.g. updating a column for every `variety` in the entire `genus`
+
+## Decision
+
+1. Hierarchy
+
+Hierarchy of entities should be implemented using foreign keys and table relationships instead of Inheritance feature of PSQL. In case of a conflict, the lowest entity level in the hierarchy is prioritized. E.g. when there is a height value in a species and in a variety, variety wins.
+
+Plant relations(e.g. like, dislike, etc.) should be implemented using separate many-to-many tables. This is because the relations are not part of the plant entity, but rather a property of the relationship between two plants.
+
```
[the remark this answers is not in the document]
aidnurs: This is a good example of what I mean in my latest comment. PSQL does not allow foreign key relations across multiple tables. To give you a more handy example: as we've already discussed that we want cross-hierarchical relations between plants e.g. a plant does not "like" an entire genus. This is not possible with metatables proposed in the decision since you cannot map multiple tables within the same foreign key.

markus2330:
> I was stuck and confused with this decision for a bit because I couldn't say for sure what the foreign key relationships will look like since I don't have enough data and insights for now. It is going to depend on the semantics of the data in the future. Let me provide you an example here:

Thank you for telling me! Please never hesitate to say that something is unclear.

> all the information scraped from practicalplants is represented in a flat form e.g. there is no semantical difference between entries yet(species vs genus vs....) i.e. we should consider them all as varieties for now.

Yvonne said that it is not so clear if there are more species or varieties in the list. The others (genus and up) cannot be in this database, as this is only an abstract concept and not a real plant in nature. Anyway, Yvonne can categorize typical plants for us so that we can get a better understanding (species or varieties; and their relations). In which format should she do it?

> I was able to extract all unique occurrences of genus and family values, however, the species list is still empty. So the relationship structure cannot be fixed yet and will take some time.

:+1: 

> I totally agree with you in your latest comment that we should keep decisions apart from technical parts, I was quite confused with the decision structure in general since I thought that a general decision would be sufficient e.g. PSQL Inheritance was rejected right away and flat table structure was preferred with the first commit. I suggest we proceed with the current state and do not fix the structure yet, so it doesn't hold our progress.

Ok, then let us keep the decision as is. Simply write in the decision what is still open.

aidnurs:
Hey @markus2330,

thank you for the response! I've updated the documentation and the decision by listing all open points. I think it is a good decision to proceed without fixing the structure yet. 

Regarding the scraper, we have all steps needed to populate the database already in place, so once we agree on the structure, we can easily adjust the required changes.

doc/decisions/database_structure.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+Each of these entities can have their own properties.
+E.g., a variety under a specie can have different height than the specie that it belongs to.
+
+The challenge is to define the structure of the database in a way that the data
+
+-   can be easily queried and
+-   updated without redundancy (e.g. setting attributes for a whole family).
```
markus2330, line 54: attribute/property used as synonym

doc/decisions/database_structure.md
```diff
@@ -0,0 +1,118 @@
...
+-   on species level.
+
+Furthermore:
+
+-   a variety belongs to a specie
+-   specie belongs to a genus
+-   genus belongs to a subfamily or family
+-   subfamily belongs to a family.
+
+Please take a look at the diagram below, which shows the hierarchy of entities with `height` and `width` as an example of attribute classification described above.
```
markus2330, line 21: attribute/property used as synonym

markus2330: we go for attributes, please add to glossary

backend/migrations/2023-02-20-213219_create_plants_metatables/up.sql
```diff
@@ -0,0 +1,34 @@
+-- Your SQL goes here
+CREATE TABLE family (
```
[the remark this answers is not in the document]
markus2330: As discussed, it should be removed.

aidnurs:
fixes #47

aidnurs:
Hey @markus2330,

this PR is ready to be merged:
- the database structure is open for future discussions as agreed yesterday
- migrations with metatables are removed
- the decision is marked as TBD in the docs
- implemented helper flags for Yvonne
- scraper/merge/insert scripts are ready. however, they could be still changed with future changes in the database structure
- documentation is up-to-date

Sincerely,
Nursultan

doc/decisions/database_structure.md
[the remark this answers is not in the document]
markus2330: Please write such outcomes (what someone said) also in the decision, so that the decision always represents the full picture of the current knowledge concerning the decision.

backend/migrations/2023-02-20-112009_alter_plant_detail/up.sql
```diff
@@ -0,0 +1,5 @@
+-- Your SQL goes here
+ALTER TABLE plant_detail
+ADD COLUMN license VARCHAR;
```
markus2330, line 3: It is better if the license is an enum, we don't want any arbitrary licenses. Are we sure that the whole database currently has the same CC license? (Did you scrape this information?)

scraper/README.md
```diff
@@ -19,20 +19,54 @@ npm install && mkdir -p data
...
 ```
 
-4. Insert data into database
+As an additional step the script will also lookup the german common name through Wikidata API and store it under the column `Column Name DE` in the `detail.csv` file.
 
-After the data is scraped and stored in `csv` format, we should now insert it into the database. This can be done with the following command:
+There will be several CSV files generated in the `data` directory:
 
-_Note:_ Please make sure that the database is up-to-date with latest migrations before running this command.
+-   `detail.csv`: This file contains the data scraped from the Practical Plants wiki dump.
```
markus2330, line 36: Please specify these files (which are CC licensed) in the file [.reuse/dep5](https://github.com/ElektraInitiative/PermaplanT/blob/master/.reuse/dep5)

scraper/README.md
```diff
@@ -19,20 +19,54 @@ npm install && mkdir -p data
...
+-   `distinctFamily.csv`: This file contains the distinct family names scraped from the Practical Plants wiki dump.
+-   `distinctGenus.csv`: This file contains the distinct genus names scraped from the Practical Plants wiki dump.
+-   `errors.csv`: This file contains the errors encountered during the scraping process. The columns in the file identify the type of error i.e. which part of the data was missing. The main error type occurs under the column "Plant Datatable" and represents listing pages e.g. [/abelia][https://practicalplants.org/wiki/abelia/].
```
markus2330, line 39: detail_yvonne.csv is nowhere mentioned? "Is Tree" is nowhere true in detail_yvonne.csv?

markus2330 (changes requested):
Good job! At least the .reuse/dep5 should be fixed before merging.

backend/migrations/2023-02-20-112009_alter_plant_detail/up.sql
aidnurs: good idea and yes, all the scraped information is available under the same license incl. the data fetched from wikipedia

aidnurs:
Hey @markus2330,

Summary:
1. As I've already mentioned we fetch the license from each practicalplants page. I've overseen the 2nd license that they use sometimes called `CC BY-NC-SA`
     > The Practical Plants database is released under a CC BY-SA license, except where otherwise specified. Some portions of the data on this website are currently under a CC BY-NC-SA license as distributed by the original copyright owner. These parts are marked with a non-commercial symbol. We are actively working to transition these portions to a CC BY-SA license.
2. Wikidata used to fetch german names uses CC BY-SA as well([link to their copyright page](https://www.wikidata.org/wiki/Wikidata:Copyright))

I have a couple of questions to clarify:
1. Should I use an array for the `license` column instead
     e.g. ["CC BY-SA", "CC BY-NC-SA"] where "CC BY-NC-SA" is coming from the practical plants and "CC BY-SA" because of the wikidata
2. Should I add only the csv files to .reuse/dep5? And how it should look like?
```
Files: scraper/src/data/*
Copyright: ___
License: CC-BY-SA-3.0, CC-BY-NC-SA-3.0
```

Thank you very much in advance,
Nursultan

markus2330:
>  Should I use an array for the license column instead
> e.g. ["CC BY-SA", "CC BY-NC-SA"] where "CC BY-NC-SA" is coming from the practical plants and "CC BY-SA" because of the wikidata

No, that would only obfuscate what the real license for each of the parts is. Let us remove the column and state at one place (.reuse/dep5) what the different licenses are. Perfectly clear (from license point of view) is if the BY-NC-SA and the BY-SA licensed content would be in different CSVs and merged only by the tool (and not checked-in). But that would break our workflow, wouldn't it?

>  Should I add only the csv files to .reuse/dep5?

Exactly.

> And how it should look like?

Depends on if we can split the files easily. If they are split it is obvious.

If it is not split, your suggestion looks fine, maybe use Copyright to write how the License must be interpreted.

```
Files: scraper/src/data/*
Copyright: different holders [or copy from original source], German names from Wikipedia (CC-BY-SA-3.0), other columns from practicalplants and Yvonne Markl (2022) (CC-BY-NC-SA-3.0)
License: CC-BY-SA-3.0, CC-BY-NC-SA-3.0
```

markus2330 (changes requested):
Great job, nearly done:

1. the build server fails for the backend (see Nextcloud talk chat)
2. detail.csv does not seem to be updated? (German name missing?)
3. add reuse license for the CSVs and completely remove the license row in the database (useless in its current form: it doesn't apply to the whole row anymore and gets ambiguous)

aidnurs:
> > Should I use an array for the license column instead
> > e.g. ["CC BY-SA", "CC BY-NC-SA"] where "CC BY-NC-SA" is coming from the practical plants and "CC BY-SA" because of the wikidata
> 
> No, that would only obfuscate what the real license for each of the parts is. Let us remove the column and state at one place (.reuse/dep5) what the different licenses are. Perfectly clear (from license point of view) is if the BY-NC-SA and the BY-SA licensed content would be in different CSVs and merged only by the tool (and not checked-in). But that would break our workflow, wouldn't it?

This would add some unnecessary complexity to the code. For example, I also create 2 additional files with distinct genus and families, so the number of files will then double in size. Yvonne would also have to work with 2 different files for each file type during her correction. However, I could then easily merge the files since each row represents different plants i.e. I just have to concat them together. So I would avoid such complications if possible, however, it is up to you how we proceed here.

markus2330:
> This would add some unnecessary complexity to the code. For example, I also create 2 additional files with distinct genus and families, so the number of files will then double in size. Yvonne would also have to work with 2 different files for each file type during her correction. However, I could then easily merge the files since each row represents different plants i.e. I just have to concat them together. So I would avoid such complications if possible, however, it is up to you how we proceed here.

I agree, simply write a "Copyright" or maybe better "Source", see https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/#source-field in .reuse/dep5 that clearly¹ writes that two licenses are mixed in one CSV. (Unless you have a better idea. I am also not a lawyer.)

¹ As written in the suggestion above: BY-SA for Wikipedia rows, BY-NC-SA for all the other rows.

kitzbergerg:
Btw it seems like https://github.com/ElektraInitiative/PermaplanT/pull/48/commits/5f1b97bf466635511efaeba1adac63d1076ee9a2 breaks something in the backend. No idea what exactly is going wrong, the errors is quite cryptic. The error occurs in `plants.rs` `find_all` method.

aidnurs:
> Btw it seems like [5f1b97b](https://github.com/ElektraInitiative/PermaplanT/commit/5f1b97bf466635511efaeba1adac63d1076ee9a2) breaks something in the backend. No idea what exactly is going wrong, the error is quite cryptic. The error occurs in `plants.rs` `find_all` method.

The problem was a missing column in the plant model. Fixed, we're good now

aidnurs:
Hey @lukashartl, could you please take a look at the CI pipelines?  In order to start the frontend, we need to rebuild the types with `typeshare ./ --lang=typescript --output-file=../frontend/src/bindings/definitions.ts`, which is actually done in `build_backend.sh` as I can see. However, during the frontend build, old types are being imported, causing pipeline failures. 

>  Module '"@/bindings/definitions"' has no exported member 'VarietyDTO'.

So in our case, we have renamed `varieties` into `plants` in the database and in the backend. As you can see from the pipeline log, the old definitions i.e. VarietyDTO(instead of PlantDTO) are being imported.

Appreciate your help, Thank you very much in advance!

aidnurs:
Hey @lukashartl,

You're totally right, thank you very much for taking a look! my bad sorry, I'm still in between the backend and frontend layers yet. Will take care of the rest

aidnurs:
Hey @markus2330,

here are the latest changes since your last review:

- I've resolved all CI pipeline issues
- german common names for plants work
- updated licenses

These were the final changes for the PR to fulfill, feel free to merge the branch once you check it.

Thank you,
Nursultan

Issue #37, opened 2023-01-27
master deployment


markus2330:
- [x] proxy setup for backend
- [x] Ansible setup to install PostgreSQL etc. (is already installed)
- [x] build&deploy [ci/Jenkinsfile](https://github.com/ElektraInitiative/PermaplanT/blob/master/ci/Jenkinsfile) (cargo run, diesel setup, diesel migration run)
- [x] make Jenkins job public (or at least accessible for PermaplanT users)
- [ ] (database) backup

markus2330:
@aidnurs @buenaflor How will we do the proxy-setup so that both the Frontend and Backend are reachable on dev.permaplant.net?

lukashartl:
Hello,

short update on the infrastructure work :rocket:

The following things have now been automized (with ansible):

* install postgres
* create postgres database + database user
* set database password
* write `.env` file
* create service user (the service will run as a user with minimal privileges)

Next up will be the [ci/Jenkinsfile](https://github.com/ElektraInitiative/PermaplanT/blob/master/ci/Jenkinsfile) as well as the whole apache2 + proxy + monit.

Expect more updates soon :rocket:

---

> @aidnurs @buenaflor How will we do the proxy-setup so that both the Frontend and Backend are reachable on dev.permaplant.net?

The following two solutions come to my mind:

1. We create an `api.dev.permaplant.net` which will only host the backend and `dev.permaplant.net` which only hosts the static (css+js) files.
2. We utilize subpaths. The Location `dev.permaplant.net/api` gets proxied to the backend, while `dev.permaplant.net` just holds the static files.

Either way, we have to inject the correct BASE_URL for backend requests into the frontend. This could happen via environment variables during build time, etc. but maybe @aidnurs @buenaflor have any better ideas :smiley:

markus2330:
Great job! :1st_place_medal: Looking forward to the PR! 🚀

lukashartl:
Hello,

next update on the infrastructure :rocket:

Dev uses the latest rust version via rustup installed to `/opt/rust`. A `source`-able env script is provided under `/opt/rust.env`. Because the `rust` version provided by the os package manager is too old (of course).

I've tried to build the latest master on dev, but unfortunately I got errors like:

```
error: Table contains more than 32 columns. Consider enabling the `64-column-tables` feature to enable diesels support for tables with more than 32 columns.
error[E0433]: failed to resolve: use of undeclared crate or module `plant_detail`
```

[...]

PR #54, opened 2023-02-19
ci: docker-compose setup (abandoned for now)

lukashartl:
Partially implements #37

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+./backend/.env
+./backend/target/
--- a/ci/backend/Dockerfile
+++ b/ci/backend/Dockerfile
@@ -1,0 +1,23 @@
+FROM rust:1.67-slim
+
+VOLUME /bindings
+
+ENV DATABASE_URL=postgres://permaplant:permaplant@127.0.0.1/permaplant
+ENV BIND_ADDRESS_HOST=127.0.0.1
+ENV BIND_ADDRESS_PORT=8080
+
+EXPOSE 8080/tcp
+
+WORKDIR /code
+RUN cargo init
+COPY backend/Cargo.toml /code/Cargo.toml
+RUN cargo fetch
+COPY ./backend/ /code
+
+RUN apt update && apt install -y libpq-dev
+
+RUN cargo install diesel_cli --no-default-features --features postgres
+RUN cargo install typeshare-cli
+
+COPY ./ci/backend/entrypoint.sh /code/entry.sh
+ENTRYPOINT /code/entry.sh
--- a/ci/backend/entrypoint.sh
+++ b/ci/backend/entrypoint.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# We need to run the migrations on all container startups
+# as the schema may have changed.
+
+# Echo on and fail fast
+set -ex
+
+# Migrate DB
+LC_ALL=C diesel migration run
+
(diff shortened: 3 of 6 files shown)


kitzbergerg:
I just ran `docker-compose up --build` and it built an ran. 

There are however a few issues for me:
* I was unable to access the App via the browser from the host. I changed the frontend entry point from `npm run dev` to `npm run dev -- --host` and then I was able to load the app from the host as well.
* I was also unable to access the backend since the port is not exposed, but adding
  ```
      ports:
        - 8080:8080
  ```
  in `docker-compose.yaml` doesn't seem to help. The backend can connect to the database, but I cannot call the backend from outside the docker-compose network which is necessary when opening the frontend in the browser.

Apart from that it might be a good idea to build the images with multiple stages. So e.g. start by building the rust binary and then use `alpine` (or another really small image) to run the executable. That way images are smaller and it might build faster. It would also be easier to split into a "build" and a "runnable" image later on if necessary (again the benefit would be that the images are way smaller as we don't need all the build dependencies).

lukashartl:
According to #37 we are focussing on #59 (which triggered #57) right now.

@kitzbergerg thank you for your feedback

kitzbergerg:
One more thing to note.

Since we are currently only in developing stages it is fine to build the rust image/binary with debug profile.  
But as soon as it is somewhat production ready/actually used somewhere we should build using release profile (`cargo build --release` instead of `cargo build` and the same goes for `cargo run --release`).

It probably not important right now, but we should be able to easily swap out those commands in the pipeline later on.

Benjaminimal:
Looks good but there is one thing that I would do differently.
I like to use `docker compose` for developing locally especially for running the db.
I work from the root dir of the project so anytime I would want to start or stop a service or execute a command in a container I'd have to `cd` into the `ci` dir or add `-f ci/docker-compose.yaml` to the command.
In the projects, I have been working with we usually put the `docker-compose.yaml` in the root dir with the `Dockerfiles` and entrypoint scripts in the respective subdirectories.
Here is an example of the layout I'm proposing.
```
├── backend
│   ├── ...
│   ├── Dockerfile
│   └── entrypoint.sh
├── frontend
│   ├── ...
│   ├── Dockerfile
│   └── entrypoint.sh
├── ...
├── docker-compose.yaml
└── Jenkinsfile
```
You would have to adjust some things in the existing files as well if you take this path.
Adjust `COPY` in the `Dockerfile`s, `context` in `docker-compose.yml`, extend `.dockerignore` not to include the `Dockerfile`, ...

markus2330:
> └── Jenkinsfile

I don't think we should put Jenkinsfile to top-level. (Btw. I generally don't like top-level files but this one docker-compose.yaml is ok.) 

But the other idea sounds like a good suggestion! Docker is only for dev, so it should be designed to be as useful as possible for dev.

@Benjaminimal do you want to take this PR over? There are also others who want to use it, so it would be good if we can get it merged while Lukas could continue to stay focused on the CI.

Benjaminimal:
@markus2330 
Sure I can take over but I'll wait for write access on this repo.
Doing this via my fork with an existing PR and branch in here will get messy as it would need a new PR.

markus2330:
You were just invited for write access and can push directly here.

Benjaminimal:
Closing this for now.
If I find time for a docker development setup I'll open a new PR.

Issue #55, opened 2023-02-20
avoid expect/unwrap at run-time


markus2330:
Currently there are a few places where expect is used in the backend (e.g.  pool.get r2d2)

These places should be moved only to startup, at run-time there shouldn't be a panic.

We also should have a look at tools like https://docs.rs/no-panic/latest/no_panic/ to explicitly mark where panic is allowed.

markus2330:
Thank you for investigating! @kitzbergerg Please do the setup for clippy. @lukashartl can then do the call of `cargo clippy` in our CI (this can actually run in Docker containers).

PR #58, opened 2023-02-23
Feature/#55 avoid unwrap expect

kitzbergerg:
Deny expect and unwrap like discussed in #55. For now allow already existing expect. Depending on whether those can actually fail we might later on remove those as well.

Fix other clippy warnings as well as they would be quite annoying when adding a clippy check to ci.

### DIFF
--- a/backend/src/services/seed_service.rs
+++ b/backend/src/services/seed_service.rs
@@ -9,9 +9,9 @@ use crate::{
     },
 };
 
-#[allow(clippy::expect_used)]
 pub fn find_all(pool: &Data<Pool>) -> Result<Vec<SeedDTO>, ServiceError> {
-    let mut conn = pool.get().expect("Failed to retrieve pool.");
+    let mut conn = pool.get()?;
+
     match Seed::find_all(&mut conn) {
         Ok(seeds) => Ok(seeds),
         Err(msg) => Err(ServiceError::new(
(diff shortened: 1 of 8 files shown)


backend/src/services/seed_service.rs
```diff
@@ -9,6 +9,7 @@ use crate::{
     },
 };
 
+#[allow(clippy::expect_used)]
 pub fn find_all(pool: &Data<Pool>) -> Result<Vec<SeedDTO>, ServiceError> {
     let mut conn = pool.get().expect("Failed to retrieve pool.");
```
markus2330, line 14: Is there maybe some way to reconnect?

markus2330 (changes requested):
Thank you for the PR! :rocket:

> Deny expect and unwrap like discussed in https://github.com/ElektraInitiative/PermaplanT/issues/55. For now allow already existing expect. Depending on whether those can actually fail we might later on remove those as well.

Please actually check (by docu etc.) if they can fail. Every `allow(clippy` should have a comment why and for which specific purpose there is this exception.

See #60.

> Fix other clippy warnings as well as they would be quite annoying when adding a clippy check to ci.

Are there still clippy warnings? If so, please create an issue (and let us keep this PR short).

kitzbergerg:
> Please actually check (by docu etc.) if they can fail. Every allow(clippy should have a comment why and for which specific purpose there is this exception.

There is now only one exception where expect is used and that is when setting up the initial connection.

> Are there still clippy warnings? If so, please create an issue (and let us keep this PR short).

I fixed them as part of this PR (there we only like 3 errors and all of just something like using return ...; instead of implicit return

markus2330:
Thank you, great job!

PR #59, opened 2023-02-23
CI: Build Stage

lukashartl:
Building frontend and backend

### DIFF
--- a/ci/build-scripts/build-backend.sh
+++ b/ci/build-scripts/build-backend.sh
@@ -1,12 +1,13 @@
 #!/bin/bash
 
 echo "building backend"
-cd ./backend/
-
-source /opt/rust.env
 
 # Echo on and fail fast
 set -ex
+
+cd ./backend/
+
+source /opt/rust.env
 
 # Download Cargo Modules
 cargo fetch
--- a/ci/build-scripts/build-backend.sh
+++ b/ci/build-scripts/build-backend.sh
@@ -3,6 +3,8 @@
 echo "building backend"
 cd ./backend/
 
+source /opt/rust.env
+
 export DATABASE_URL=postgres://cidb:cidb@127.0.0.1/cidb
 
 # Echo on and fail fast
--- a/ci/build-scripts/build-backend.sh
+++ b/ci/build-scripts/build-backend.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+echo "building backend"
+cd ./backend/
+
+export DATABASE_URL=postgres://cidb:cidb@127.0.0.1/cidb
+
+# Echo on and fail fast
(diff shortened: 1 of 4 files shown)


ci/build-scripts/build-backend.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+echo "building backend"
+cd ./backend/
+
+source /opt/rust.env
```
markus2330, line 6: Why not fail here on errors?

ci/build-scripts/build-backend.sh
```diff
@@ -0,0 +1,25 @@
...
+
+export DATABASE_URL=postgres://cidb:cidb@127.0.0.1/cidb
```
markus2330, line 8: What about setting this in ci/Jenkinsfile to keep the script more useful for devs? https://www.jenkins.io/doc/pipeline/tour/environment/

ci/build-scripts/build-backend.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Echo on and fail fast
+set -ex
+
+# Reset Database
+sudo /usr/local/bin/permaplant-reset-ci.sh
```
markus2330, line 14: Same here: maybe this can be done outside. It should be also clear at which pipelines it is done (at least write this in a comment).

ci/build-scripts/build-frontend.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+echo "building frontend"
+cd ./frontend/
+
+# Echo on and fail fast
+set -ex
```
markus2330, line 7:
```suggestion
echo "building frontend"

# Echo on and fail fast
set -ex

cd ./frontend/
```

markus2330 (changes requested):
Please add dev documentation how CI works, which should also clarify when database gets reset.

Otherwise LGTM, maybe it is possible to have the scripts also working outside CI context (if not, it is fine, too).

lukashartl:
/cc @markus2330 
Updated. Can be merged. Please `Squash and merge`.
I'll create a new PR for the Dev-Instance, so we can at least do the test builds on all PRs from now :)

markus2330 (approved):
LGTM, why do you want it squashed? Is there something within the commits which shouldn't be published? If not, I would prefer to have original commits. They are sometimes insightful how/why something was done.

lukashartl:
> LGTM, why do you want it squashed? Is there something within the commits which shouldn't be published? If not, I would prefer to have original commits. They are sometimes insightful how/why something was done.

I am ok with either ways :) 

I just thought that we don't need that many "trying things out"-commits.

PR #62, opened 2023-02-26
Add contribution/docu guidelines

markus2330:


aidnurs (approved):
Introducing templates is such a nice idea, I would also suggest taking `Basics` section out into its own `Developing Guidelines` page. As an example, some essential points for new developers from my experience:
- who is merging the final PR
- commit message structure
- why/when to use decisions
- etc.

markus2330:
@kitzbergerg please have a look if the branch names are okay (if not, please suggest how)

absurd-turtle (approved):

Issue #65, opened 2023-02-26
backend architecture


markus2330:
We should have full architecture description of how the code in the backend should be structured.

Please use arc42 and a similar style as already started (and also used in Elektra) with Markdown files per chapter.

markus2330:
@kitzbergerg Did you already read through the whole backend code? Are there obvious improvements?

kitzbergerg:
I haven't looked into it in detail, but I noticed a few things.
One thing I remember is:  We should use `From<Error>` or `.map_err()` more often to convert errors and then use `?`. There are quite a few long and somewhat unreadable match statements that could be removed with changes like that.

How do we want to handle that? Can I just start and change the code how I think it would make sense and book it to either this ticket or #60. While doing so I would also create a style guide. Although I'll try to mostly use clippy lints.

For a higher level overview we could use <https://github.com/rust-lang/mdBook> and for code documentation `cargo doc`. Both generate webpages that look in my opinion quite nice (mdBook is used for the [rust book](https://doc.rust-lang.org/book/) and `cargo doc` is what is used in crates.io)

markus2330:
"Obvious" improvements (like making code shorter/more readable/...) you can directly do (simply create PR(s)). For bigger restructuring it is needed to first create architecture/docu/decisions so that I can give my okay first.

> For a higher level overview we could use https://github.com/rust-lang/mdBook and for code documentation cargo doc. 

Sounds good to me! :rocket:

kitzbergerg:
@markus2330 Is there a reason why we use routes instead of macros (see <https://actix.rs/docs/getting-started>). Otherwise I'll refactor that as well in a sensible way.

markus2330:
I don't think there is a special reason for this. (@buenaflor?)

Feel free to refactor if it leads to easier comprehensible code.

kitzbergerg:
Do we want path normalization? (See <https://actix.rs/docs/url-dispatch#path-normalization-and-redirecting-to-slash-appended-routes>)

Currently there we get 404 if we call e.g. `http://localhost:8080/api/seeds/`.  
With normalization both `http://localhost:8080/api/seeds` and `http://localhost:8080/api/seeds/` would match (as well as e.g. `http://localhost:8080/api//seeds/`)

markus2330:
@kitzbergerg  If it doesn't cost performance or doesn't require much additional code it is handy for manual testing (when using curl etc.).

kitzbergerg:
[...]

Issue #60, opened 2023-02-24
coding guidelines


markus2330:
We should write a document similar to https://github.com/ElektraInitiative/libelektra/blob/master/doc/CODING.md which describes a common style to be checked in reviews. As much as possible should already be checked by linters, so that the document can be short.

As first rule I suggest, that breaking any rules (like not using expect at run-time) is only allowed if a code comment argues why (e.g. "expect is okay here because the string is static and parsing JSON from it can never fail").

PR #68, opened 2023-02-28
Backend code refactoring

kitzbergerg:
A start to #60 and #65.

### DIFF
--- a/backend/Cargo.lock
+++ b/backend/Cargo.lock
@@ -261,6 +261,7 @@ dependencies = [
  "actix-rt",
  "actix-web",
  "chrono",
+ "derive_more",
  "diesel",
  "diesel-derive-enum",
  "dotenvy",
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -9,6 +9,7 @@ actix-web = "4"
 actix-rt = "2.8.0"
 dotenvy = "0.15.0"
 typeshare = "1.0.0"
+derive_more = "0.99.17"
 
 [dependencies.diesel]
 version = "2.0.2"
--- a/backend/src/config/db.rs
+++ b/backend/src/config/db.rs
@@ -9,10 +9,8 @@ pub type Pool = r2d2::Pool<ConnectionManager<Connection>>;
 pub fn config(url: &str) -> Pool {
     let manager = ConnectionManager::<Connection>::new(url);
 
-    // Panics if the connection to the database could not be established after a certain period of time.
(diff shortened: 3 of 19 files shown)


kitzbergerg:
@markus2330 Is it ok to change the error response in the backend? Previously `ResponseBody` was used for errors as well. That however has no benefit at all as `data` was always empty. In this PR I change the response to be just a string. If this is a problem I'll change it back.

Generally I would prefer to not use `ResponseBody` as the `message` part of that struct is mostly unnecessary in my opinion. Currently the only thing that is sent in `message` is `ok`, but we can just use the status code to convey that the request was successful.
This would however require a change in the frontend as well so I don't know how to handle that.

markus2330:
> Is it ok to change the error response in the backend?

Yes, totally. Looking forward to see a nice error concept by a Rust expert! :1st_place_medal: 

Can you write a small technical document on how to handle errors in PermaplanT for Rust novices?

> mostly unnecessary

When is it necessary? Is it worth to have differences in the APIs in these situations?

> This would however require a change in the frontend as well so I don't know how to handle that.

Due to typeshare+typescript it will give you compile errors.

kitzbergerg:
> Can you write a small technical document on how to handle errors in PermaplanT for Rust novices?

I would make that part of the high level doc using `mdBook`.

> > mostly unnecessary
> 
> When is it necessary? Is it worth to have differences in the APIs in these situations?
> 
> > This would however require a change in the frontend as well so I don't know how to handle that.
> 
> Due to typeshare+typescript it will give you compile errors.

Maybe create a separate issue for that.

markus2330 (approved):
Thank you for the improvements!

In general we are missing code comments. Most of the devs are new to Rust, so some explanations of the code are useful.

> I would make that part of the high level doc using mdBook.

:+1: This should be the next step.

>   Maybe create a separate issue for that.

Or a separate PR so that we can see the compile error. But first we should clarify the architecture, so that we don't need to go back and forth.

kitzbergerg:
> In general we are missing code comments. Most of the devs are new to Rust, so some explanations of the code are useful.

This will be part of #60 (which I'm already working on). Part of that forces you to write doc for every method in the project.

markus2330:
Please don't self-merge, see also #62. But I was just waiting for your answer and would have merged then, so nothing happened.

kitzbergerg:
Oh sorry about that then. You approved the request so I thought it would be fine.

PR #71, opened 2023-03-01
Meeting 02.03.2023

aidnurs:
Hey team,

please let me know what you think about such a structure for the meetings, where we submit the 3 points, that we would like to discuss here in GitHub, along with the current state(optional) of our works. The main thought behind this was to get involved in the work of others and have a better overview of the project. For example, I never read the issues and PRs where I'm not mentioned or assigned.

Please note that this is just an idea and could be a slight overkill. As always I'm open to your feedback!

Sincerely,
Nursultan


markus2330 (approved):
> please let me know what you think about such a structure for the meetings, where we submit the 3 points, that we would like to discuss here in GitHub, along with the current state(optional) of our works.

I like the idea very much! Can you also add a template to copy for every meeting?

> The main thought behind this was to get involved in the work of others and have a better overview of the project. For example, I never read the issues and PRs where I'm not mentioned or assigned.

Yes, absolutely. It is important that everyone is aware of what others are working on.

> Please note that this is just an idea and could be a slight overkill. As always I'm open to your feedback!

We should decide to either:
- simply have textfile without folders (simpler)
- don't add links to the individual files (then the PRs would be conflict-free). The main file is better called README.md (to not repeat the date).

absurd-turtle (approved):

PR #72, opened 2023-03-01
coding guidelines

kitzbergerg:
As part of #60 I want to enable all clippy lints that seem useful. 

This also includes adding documentation to all methods and modules in the backend.

After all clippy lints are active there might not even be a need to write additional coding guidelines (as clippy can be quite strict).


kitzbergerg:
@markus2330 Please have a lot at the metadata I added to the toml file. I put in there what clippy wanted and tried to piece together what I think would make sense.

backend/src/main.rs
```diff
@@ -1,6 +1,60 @@
...
+    clippy::unnecessary_self_imports,
+    clippy::unneeded_field_pattern,
+    clippy::unreachable,
+    clippy::unseparated_literal_suffix,
+    clippy::unwrap_in_result,
+    clippy::unwrap_used,
+    clippy::use_debug,
+    clippy::verbose_file_reads
+)]
+// Cannot fix some errors because dependecies import them.
```
markus2330, line 46: Please write which ones are effected (so that we might reenable if the dep is gone).

markus2330 (changes requested):
Amazing investigation! Really like the many lints :heart: 

Only two very minor points, see comments.

Some of them might be too much, e.g. partial_pub_fields, but lets see if we hit them.

backend/src/main.rs
kitzbergerg: In this specific case it is `time: 0.1.45, 0.3.17` and `wasi: 0.10.0+wasi-snapshot-preview1, 0.11.0+wasi-snapshot-preview1`. Both of them are sub-dependencies of other crates. So I would keep `#![allow(clippy::multiple_crate_versions)]` as we would never really know what might happen when we add new crates.

backend/src/main.rs
```diff
@@ -1,6 +1,60 @@
...
+#![warn(clippy::nursery)]
+#![warn(clippy::cargo)]
+// Lints in clippy::restriction which seem useful.
+#![warn(
+    clippy::clone_on_ref_ptr,
+    clippy::empty_structs_with_brackets,
+    clippy::exit,
+    clippy::expect_used,
+    clippy::format_push_string,
+    clippy::get_unwrap,
```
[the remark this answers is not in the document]
kitzbergerg: There is [indexing_slicing](https://rust-lang.github.io/rust-clippy/master/#indexing_slicing).
I will add that, then in combination with `get_unwrap` (or simply the normal unwrap one) only `.get()` and then handling the `None` is allowed.

backend/src/main.rs
markus2330: Actually I think that in such cases most likely the crates should update their dependencies. So if multiple_crate_versions would be used by many applications, libs depending on old libs would probably happen less. But of course we cannot influence it directly and during dev it would be rather disturbing, so let us close this topic for now.

backend/src/main.rs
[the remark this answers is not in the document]
markus2330: Perfect!

kitzbergerg:
I'm kind of finished with what I can do in this PR for now (not yet to be merged). 

I documented where I could in code and left TODOs where I didn't really know what to write (e.g. struct fields that represent the database). For that I would need some good documentation or some help to tell me what the individual fields are actually for (maybe discuss that on Tuesday or somebody helps me after the meeting).

Then there is still `#![allow(clippy::module_name_repetitions)]`. I don't actually know whether we want that, but I can't really do anything about that currently as I'm kind of blocked from renaming file by #48 (we did some renaming there and that would result in a huge merge conflict mess).

kitzbergerg:
@markus2330 To view the docs type `cargo doc --open` and the generated html files should open in the browser. It's easier to look through code that way.

markus2330 (changes requested):
Great job! :heart: 

- please also add `cargo doc --open` to the documentation (README.md
- struct fields that represent the database: Yes, good idea to document it here. #48 and others can add documentation there.
- module_name_repetitions: I am all for not repeating module names. Not having to do that, is one of the main reasons why modules exists.

kitzbergerg:
@lukashartl Once this PR is merge you probably need to change CI. There should be an additional check whether documentation links are broken. The command for that is `cargo doc`.

@markus2330 It is possible to set a flag in rust to deny warnings (https://github.com/rust-lang/cargo/issues/8424#issuecomment-774662296). We might want to enable that.

markus2330:
> @lukashartl Once this PR is merge you probably need to change CI. There should be an additional check whether documentation links are broken. The command for that is cargo doc.

> @markus2330 It is possible to set a flag in rust to deny warnings (https://github.com/rust-lang/cargo/issues/8424#issuecomment-774662296). We might want to enable that.

I created #98 to track this, the deployment is higher priority though.

markus2330:
Great work! Now please concentrate on documentation, architecture, big picture of backend etc., so that we can finish the first milestone without bigger merge conflicts.

kitzbergerg:
> Great work! Now please concentrate on documentation, architecture, big picture of backend etc., so that we can finish the first milestone without bigger merge conflicts.

There are still a few smaller issues open. I will quickly fix them and afterwards start with the doc/decisions.

Issue #35, opened 2023-01-26
seeds entry


markus2330:
- [ ] Sorte: auto-selection should be based on Art chosen before
- [ ] Sorte: it should be possible to also enter own text
- [ ] english/german mixture (should be everything English with translations)
- [ ] date: it should be also possible to only enter year (if day/month is unknown) Otherwise date entry is very fancy :wink: 
- [ ] should have proper overview of seeds entered in http://localhost:5173/seeds (and changeable)
- [ ] immediately pressing "Abbrechen" doesn't need confirmation
- [ ] generation: negative numbers should be disallowed
- [ ] browser back button and then abort: cannot press yes
- [ ] via network I got when going to "New Entry" "AxiosError: Network Error" (probably it tries to connect to wrong backend 127.0.0.1? I tried with `npm run dev -- --host 0.0.0.0`)
- [ ] http://x.x.x.x:5173/seeds -> no network error but it also does not list seeds

We entered:
```json
{"message":"ok","data":[{"id":1,"name":"Tomate","variety_id":1,"harvest_year":2023,"quantity":"Enough","tags":["Fruit crops"],"use_by":null,"origin":null,"taste":null,"yield_":null,"generation":null,"quality":null,"price":null,"notes":null},{"id":2,"name":"Tomate","variety_id":1,"harvest_year":2022,"quantity":"Enough","tags":["Fruit crops"],"use_by":null,"origin":"Arche Noah","taste":null,"yield_":null,"generation":null,"quality":"Organic","price":null,"notes":"Für Salat und zum Kochen geeignet."},{"id":3,"name":"Mangold","variety_id":1,"harvest_year":2021,"quantity":"More than enough","tags":["Leaf crops"],"use_by":"2025-01-26","origin":"Stanzach","taste":"lecker :-)","yield_":"1","generation":-2,"quality":"Not organic","price":null,"notes":""},{"id":4,"name":"Spinat","variety_id":1,"harvest_year":2018,"quantity":"Not enough","tags":["Leaf crops"],"use_by":"2021-01-01","origin":"Kiepenkerl","taste":null,"yield_":null,"generation":0,"quality":"Organic","price":null,"notes":"schnellwachsend & widerstandsfähig"}]}
```

buenaflor:
>  Sorte: auto-selection should be based on Art chosen before
[...]

PR #77, opened 2023-03-05
Remove kategorie from seed entry and convert variety to a simple text field.

badnames:
This pull requests implements the following changes requested by @markus2330 in #35:

> completely remove "Kategorie" (seeds and plants database). So the very first entry field should be Art, then Sorte, then Year.

> Variety, i.e., Sorte later will be select-able from database but for the first milestone it is simply text to be entered (As I assume we will not be able to implement the "Edit Plants" use case next week, which would be needed to add the varieties.)


backend/migrations/2023-03-03-140707_convert_variety_to_text/up.sql
```diff
@@ -0,0 +1,10 @@
+-- Your SQL goes here
+ALTER TABLE seeds DROP COLUMN variety_id;
+DROP TABLE varieties;
+
+-- Add column without not null constraint.
+ALTER TABLE seeds ADD COLUMN variety TEXT;
+-- Then set all varieties to some default value .... 
+UPDATE seeds SET variety = 'Default';
+-- ... before setting not null constraint.
+ALTER TABLE seeds ALTER COLUMN variety SET NOT NULL;
```
markus2330, line 10: Variety of seeds can be NULL. This represents that the the variety of the plant_id is used.

markus2330 (changes requested):
Great job! Thank you for the cleanup!

- please update the README.md from backend (there is no varieties anymore)
- http://127.0.0.1:5173/seeds nor curl 192.168.178.24:8080/api/seeds shows the seeds I just entered? Is it really inserted in the DB? This worked before.

badnames:
Thank you for the review @markus2330.

> please update the README.md from backend (there is no varieties anymore)
Yes I missed that. Its now fixed. 

> http://127.0.0.1:5173/seeds nor curl 192.168.178.24:8080/api/seeds shows the seeds I just entered? Is it really inserted in the DB? This worked before.

I could not replicate this on my machine. Requests from the frontend, as well as plain http requests sent using curl resulted in new database entries.
Perhaps there was a bug somewhere that was now fixed by accident?

Please verify that this issue is still present with the latest changes and send the relevant curl command if the issue persists.

> Documentation/frontend-tests could be improved

I agree, but I would prefer to open another pull request for this issue as it is unrelated to #35 in my opinion.

markus2330:
> I could not replicate this on my machine.

My suspicion is that my backend was wrongly configured. Afaik the frontend does not give an error if the connection fails, so I did not notice. I added this to #35.

We now merged the PR with the new plant table (replacing varieties). This created quite a few conflicts here. Let us discuss today in which order we will finish the first milestone.

badnames:
I was just resolving the merge conflicts and encountered a problem.

As it is, both the selction field for plants and for variety is called "Sorte". 
How should these textfields be renamend?

badnames:
I have noticed that the new seed property plant_id from the master branch is also called "Sorte" now. 
How should we rename  the variety field to distinguish these two properties?

badnames:
Ok, thank you for the answer.
I merged the changes from master into this branch by using two text fields for now.

After the build issues are fixed, I will look into merging both text fields as requested.

badnames:
> There is a double use for Sorte (German)/variety (English) field:

>    if entered in the Seed entry manually, the Seed database gets an entry variety and this is preferred
>    the "Plant" is a variety, then the data is used from the Plant database (variety in seed database is NULL in this case)

Users now get a single selectbox/typeahead that allows them to create items that are not in the list. If a new item is created, it will be stored in the variety field of the seed table, otherwise the seed_id will be set as before.

markus2330:
@lukashartl do you know when we will be able to deploy PRs? It would be handy here if I don't need to test locally.

markus2330:
Great job! Looking forward to other fixes of #35.

PR #81, opened 2023-03-07
Fix: German common name fetch from Wikipedia

aidnurs:
solves #47


markus2330:
Thank you, great job! @yvonnemarkl will have a look!

PR #86, opened 2023-03-09
Clarify the use case for the plants layer

aidnurs:


doc/usecases/layers/plants_layer.md
```diff
@@ -2,36 +2,36 @@
...
+-   **Main success scenario:**
+    -   The user is able to add plants and is presented with a list of seasonal plants (for the selected time of the year), in following groups:
+        -   available seeds with early expiration days first
+        -   diversity constraints: suggestions which not-yet-used plants fit next to the plant just added, or at a place selected (spatial and temporal)
+        -   which were recently planted
+        -   and lastly favourites of users
+    -   The user gets a visual indication which parts of the map are better suited for the plant to be placed (e.g. zones, pH value, moisture, animals or shadows).
+    -   The user positions the plant element in the appropriate location in the map.
+    -   When drawing an area of plants, the size of the arena and the number of plants is shown next to the mouse.
+    -   The user is able to view the relationships between the plants by looking at the lines connecting them.
```
aidnurs, line 24: Let's specify the "lines" between entities

aidnurs, line 24: Let's specify the "lines" between plants mentioned in this use case i.e. how they should be presented.

There are the following cases that I see right now:

1. If the lines are drawn just on top of the map between plants while the plants layer is enabled
2. If the lines are drawn after the user clicks a special button to indicate the relationships, lines are also drawn on top of the map.
3. After the user clicks on a specific plant to indicate the relationships relative to the chosen one

Could you please share how do you see this feature? Let's update the description here

@markus2330 
@buenaflor

markus2330: Excellent questions, I created #95 for clarifications. Please comment there.

>  If the lines are drawn just on top of the map between plants while the plants layer is enabled

They are only drawn while drag&drop/move the plant, which is only possible if you are in the plant layer, then the lines get added to the warnings layer.

> If the lines are drawn after the user clicks a special button to indicate the relationships, lines are also drawn on top of the map.

This "special button" is the visibility of the warnings layer.

> After the user clicks on a specific plant to indicate the relationships relative to the chosen one

Yes, while moving plants, the relationships etc. should be shown again.

markus2330 (commented):
Is it only changing of whitespace? I am okay either way with whitespace. Best would be if prettier simply does it for us.

aidnurs:
Hey @markus2330,

> Is it only changing of whitespace? I am okay either way with whitespace. Best would be if prettier simply does it for us.

My original idea was to open a PR with some initial changes to start working on the clarifications through GitHub suggestions together. Since you've extended the clarifications with more information in an independent PR, we can close this one. Thank you for your quick response and contribution!

PR #92, opened 2023-03-10
Show warning modal only when the form was changed.

badnames:
Make the cancel modal only show up if the user has made form inputs (as mentioned in #35).
For this purpose, all existing form components got an onChange callback, that fires every time the user makes an input.

### DIFF
--- a/frontend/src/components/Form/CreatableSelectMenu.tsx
+++ b/frontend/src/components/Form/CreatableSelectMenu.tsx
@@ -13,6 +13,7 @@ export interface CreatableSelectMenuProps<T extends FieldValues> {
   placeholder?: string;
   handleOptionsChange?: (option: any) => void;
   handleCreate?: (inputValue: string) => void;
+  onChange?: () => void;
 }
 
 export default function CreatableSelectMenu<T extends FieldValues>({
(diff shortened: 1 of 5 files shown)


markus2330 (approved):
Works for me! A few code comments+tests would be great!

badnames:
> A few code comments+tests would be great!
@markus2330 which newly added code passages need more comments in your opinion?
Regarding test: if required, I could look into adding UI tests as part of another PR.

markus2330:
>   @markus2330 which newly added code passages need more comments in your opinion?

The comment was not specific to your code. Its only that now you dive in the code, you have a fresher perspective, which might lead to better documentation, than you will have in a few weeks.

> Regarding tests: I could look into adding UI tests as part of another PR.

I was mainly thinking about unit tests. @aidnurs was also interested in UI tests. We first need a decision about which UI test framework we use.

badnames:
> The comment was not specific to your code. Its only that now you dive in the code, you have a fresher perspective, which > might lead to better documentation, than you will have in a few weeks.

Ok I understand. I will look into documenting the frontend either today or tomorrow.

PR #93, opened 2023-03-10
Dissalow negative numbers in generation field.

badnames:
As mentioned in #35, a plants generation should not be a negative number.
For this purpose, I added min and max attributes to SimpleFormInput and applied them to the relevant input field on the seed entry page.

### DIFF
--- a/frontend/src/components/Form/SimpleFormInput.tsx
+++ b/frontend/src/components/Form/SimpleFormInput.tsx
@@ -9,6 +9,8 @@ interface SimpleFormInputProps<T extends FieldValues> {
   isArea?: boolean;
   required?: boolean;
   type?: HTMLInputTypeAttribute;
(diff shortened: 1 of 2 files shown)


markus2330 (approved):
Well done, works for me!

Writing -1 in the field still works (and is only prohibited when inserting). I assume it would be not so easy to have while-typing-validation?

badnames:
Yes it is possible to enter a negative number manually. However, the user will not be able to submit the form if a negative number is entered.

If required, I could try to a) display the error message immediately display the error message after a negative input was received, or b) automatically reset the number field to 0.

markus2330:
> If required, I could try to a) display the error message immediately display the error message after a negative input was received, or b) automatically reset the number field to 0.

No, status quo is better.

Issue #91, opened 2023-03-10
feasibility: undo-redo functionality of the map


aidnurs:
## Goal 
The main goal of this task is to understand how to store the state of the map in the database e.g. 5 red colored rectangles are drawn on the map at position XY. 

## Outcomes
The canvas library, that we use, provides a suggestion on how to implement undo-redo functionality with the canvas.

https://konvajs.org/docs/react/Undo-Redo.html
https://konvajs.org/docs/data_and_serialization/Best_Practices.html#page-title

> If you want to save/load simple canvas content you can use the built-in Konva methods: node.toJSON() and Node.create(json). But those methods are useful only in very small apps. In bigger apps it is VERY hard to use those methods. Why? Because the tree structure is usually very complex in larger apps, you may have a lot of event listeners, images, filters, etc. That data is not serializable into JSON (or it is very hard to do that).

In other words, the Konva nodes contain too much information e.g. color or shape which could be set by default. So there is no reason to store them in the database if they are set by default by Konva anyways.

> You just need to save a history of all the state changes within your app. There are many ways to do this. It may be simpler do to that if you use immutable structures.

The statement above should look the following way in our app:
1. we have a simple rectangle of a default size and color at position (20,20) saved in the database from the user's previous session
3. we rebuild the canvas based on the information from (1)
4. canvas is ready to be used
5. user moves it to a new position (40,20)
6. we save the new state of the rectangle to the `state` array, which updates the database

The main drawback of the method is that the structure of the state will be hard to debug and maintain on the frontend side e.g.
```
state[
     {
          timestamp:1,
          root{
              rootProp1,
              rootProp2,
              rootProp3,
              ....
              children:[
                    ...
              ]
           }
     },
     {
          timestamp:2,
          root{
              rootProp1,
              rootProp2,
              rootProp3,
              ....
              children:[
                ....
              ]
           }
     },
]
```

However, on the database level, such a state could be stored in a deserialized way using `JSON.stringify(state)` i.e. convert a root object with all its children into string

markus2330:
Great investigation, thank you!

> The main goal of this task is to understand how to store the state of the map in the database

:+1: Super important!

> undo-redo functionality

Isn't this more or less independent if we implement undo in the database? What about  getting everything new from the backend+redraw from scratch for undo/import/changing layers etc? Would this be too slow?

[...]

PR #94, opened 2023-03-10
Feasibility poc

aidnurs:
Solves #91 

Check out the README.md file in the `benchmarks` folder for a detailed explanation


markus2330:
jenkins build please

markus2330:
jenkins build please

markus2330:
jenkins build please

aidnurs:
Hey @kitzbergerg @markus2330 ,

could you please help me out here? I'm building a POC bzw. feasibility study for the preloading feature + undo/redo of the map, so I created a `map` table in SQL with a JSONB column storing all of the information related to the canvas objects i.e. shapes, coordinates, etc. JSONB is a temporary solution of course, it is simple and fast to implement.

So I also built the backend, but there is a problem with the JSONB type in rust. I don't have to look at the entire PR but only the backend files, which are built the standard way. It would be great if you could assist me here, I just need two simple get queries(find by id, find all) to proceed.

Thank you very much!
Nursultan

kitzbergerg:
@aidnurs Do you know what the value of `entity::Map.detail` will be? Or is it just some object that might change arbitrarily?

If `detail` has a fixed schema you should create a second struct and then use that instead of Value e.g.:
```rust
struct MapDetail {
  inner_detail: String,
  number: u32
}

pub struct Map {
    pub id: i32,
    pub detail: MapDetail,
    pub created_at: chrono::NaiveDateTime,
    pub updated_at: chrono::NaiveDateTime,
}
```


If `detail` changes each time you call the endpoint (so if the JSON has different keys each time) `Value` should be fine.

markus2330:
Great job! A few things that disturb/confuse me:

- The objects inserted before behave differently: when moving them they are still displayed at the original spot.
- Hide rect layer only works once?
- Sometimes the circles hide when trying to hide rects.
- when moving a circle, also a rect gets moved?
- undo also works on hiding rects?
- reset zoom would be useful
- Probably a counter how many circles and rects are (visible and invisible) on the map would be helpful, too.
- Please invert zoom direction :wink: 

I am happy with undo/redo. It is probably really not a big deal that it gets lost after reload. As long as there is save/load. But it seems to not always load what was saved last when refreshing the browser?

Collaboration does not work at all, as the save seems to be a quite fast operation it shouldn't hurt do it with every action? Can we get updates of what happened at the server side or would that be too much for the POC?

I need a README.md to understand what you implemented, tested and what your results are. How many rectangles got added, how many circles etc.?

aidnurs:
Hey @markus2330,

thank you for your comments! The PR contains not only an updated and working version of the canvas but also automated benchmarking. As I've updated the PR description, check out the README.md in the benchmarks folder

benchmarks/README.md
```diff
@@ -0,0 +1,99 @@
...
+1. Install dependencies
+
+```shell
+npm install
+```
+
+2. Populate the database with example maps
+
+In order to benchmark the performance of the canvas, we need to populate the database with example maps.
+There are maps of different sizes and complexity, which are stored in a single CSV file called `maps.csv` in the `/benchmarks` folder.
```
markus2330, line 19: There is no such file, only map.csv exists.

benchmarks/README.md
```diff
@@ -0,0 +1,99 @@
...
+```shell
+npm run dev
+```
+
+5. Run benchmarks
+
+```shell
+npm run benchmark
+```
+
```
markus2330, line 46: Also describe how to manually inspect the tests (maybe in other readme, so that benchmarks/ can eventually be merged). When I open e.g. http://localhost:5173/demo/6 it only shows an empty canvas. When I press "Generate" it only generates 1000 elements (and not 10000 as I would have expected from this test).

benchmarks/README.md
```diff
@@ -0,0 +1,99 @@
...
+-   `<test_name>-report.json` - the raw report of LightHouse for a single test case (e.g. `100 circles` on the canvass)
+-   `<timestamp>-lighthouse-results.csv` - the results of the audit for all test cases (e.g. `100 circles`, `1000 circles`, `10000 circles` etc.)
+-   `<timestamp>-lighthouse-results-shapes.csv` - the results of the audit for all test cases related to shapes (e.g. `100 circles` vs. `100 stars`).
+
+Metrics:
+
+-   `First Contentful Paint (FCP)` - the time it takes for the browser to render the first bit of content on the page. Measured in milliseconds.
+-   `Interactive` - the time it takes for the page to become fully interactive. Measured in milliseconds.
+-   `Score` - the performance score of the page. Measured in percentage. The metric is calculated based on the weighted average of the other metrics. (PENDING, will probably be removed because it depends on the other metrics, which are not considered in the study)
```
markus2330, line 82: Where do I see it? Which values did you get? In the playright report I have e.g.:

```
10000 circles with 10 layer
12msBefore Hooks
23msbrowser.newPage— canvas_performance.spec.js:15
20mspage.goto(http://localhost:5173/demo/7)— canvas_performance.spec.js:16
29mspage.evaluate— ../node_modules/playwright-lighthouse/src/audit.js:40
35msAfter Hooks
```
stdout:

```
Starting Lighthouse audit for 10000 circles with 10 layer


-------- playwright lighthouse audit reports --------


performance record is 62 and desired threshold was 0
Lighthouse report saved to report/10000 circles with 10 layer-report.json
Lighthouse results saved to report/1680358909532-lighthouse-results.csv
```

benchmarks/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "version": "1.0.0",
+    "description": "",
+    "main": "index.js",
+    "scripts": {
+        "benchmark": "playwright test",
+        "start": "node insert_maps.js"
+    },
+    "keywords": [],
+    "author": "",
+    "license": "ISC",
```
markus2330, line 12:
```suggestion
    "license": "BSD",
```

markus2330 (changes requested):
Good job, looks already much better. Please carefully separate benchmarks/ to be merged later and the rest (not to be merged). Benchmark results should be in a separate PR, where you also write your conclusion (how the architecture should look like and how to use the canvas etc.).

Where do I see the goals you have for the feasibility study? This should be clarified first.

benchmarks/README.md
aidnurs: The outputs are well defined just above the lines you commented on:

> The audit measures the performance of a web page by generating a performance score and then saves the results of the audit in `report` folder:

I'm not sure why you've been looking in another folder, maybe just overseen the lines. Let me know if it was still unclear to you where to find the results and what is the structure of the benchmarking. And I will change the README based on your suggestion

https://github.com/ElektraInitiative/PermaplanT/pull/94/files/d113eeec53f081a34bda0c7beea5cc90ff459cc4#diff-576c27794bae284efc278336a16d0649aaf599973a64d4b65dc67bc99bb7f000R72

benchmarks/README.md
aidnurs: I skipped an installation step populating the database with maps in the README folder, was it on purpose or by accident? Let me know how could I update the README to indicate its strictness

benchmarks/README.md
markus2330: Probably there is just some problem with terminology. What is the difference between "performance score", "performance" (as in the*-lighthouse-results.csv) and "score" (as defined in the commented line)? Anyway, you should first clarify for yourself which score actually is the most useful and better describe what it is.

PR #95, opened 2023-03-10
rework/expanding on features and clarifications

markus2330:
## Basics

## Checklist

      (not in the PR description)

## Review

- [x] Documentation is conforming to [our Documentation Guidelines](/doc/documentation.md)
- [x] Examples are well chosen and understandable
- [x] Code is conforming to our Coding Guidelines


doc/usecases/layers_alternatives.md
```diff
@@ -0,0 +1,30 @@
+# Use Case: Layer Alternatives
```
aidnurs, line 1: I prefer the "versions" as a name for the feature because it would semantically mean that I could build and revert something incrementally like migrations. However, this is just a side note and is not a strong preference of mine compared to the actual implementation details

aidnurs (approved):

doc/usecases/layers_alternatives.md
aidnurs: In general, everything looks sharp and could be merged. Thank you once again!

markus2330: > I prefer the "versions" as a name for the feature because it would semantically mean that I could build and revert something incrementally like migrations. However, this is just a side note and is not a strong preference of mine compared to the actual implementation details

I think there is a misunderstanding of this feature. (Your term definition of versions gave me the indication about this misunderstanding.) I agree with you about the term: versions usually describe linear history. But in this usecase it is *not* about linear history.

If you don't like the term "alternatives", other fitting words would be scenarios, branches, forks, ...

But we need two different terms, as we have two different concepts here.

PR #101, opened 2023-03-12
add swagger doc using utoipa resolves #85

### LINKED ISSUE #85: Swagger doc for the backend
I just found https://github.com/juhaku/utoipa/tree/master. It's able to automatically generate OpenApi doc for structs and endpoints can be annotated with macros. 
According to the example it is also able to create a swagger endpoint (see https://github.com/juhaku/utoipa/blob/master/examples/actix-web-multiple-api-docs-with-scopes/src/main.rs#L27-L36)

kitzbergerg:
## Basics

## Checklist

      (not in the PR description)

## Review


kitzbergerg:
@markus2330 here is how the swagger doc using utoipa as described in #85 would look like. Currently there is quite a lot of configuration that has to happen and generics are only somewhat nice to use (e.g. the get endpoint for seeds doesn't display an array in swagger, because it is a double nested generic `Body<Vec<SeedDto>>`).

An easy solution for that would be to remove the `Body<T>` struct as mentioned here: https://github.com/ElektraInitiative/PermaplanT/pull/68#issuecomment-1448299425. It shouldn't take longer that ~30min to remove body (including the frontend).

markus2330:
Please also write a decision about it so that it is clear what the pros/cons are. In particular it should be clear which extra effort it is to maintain it.

kitzbergerg:
@markus2330 I just pushed the decision. Is that what you are looking for?

If that is ok I would continue with Utoipa and as a next step remove the `Body` struct from frontend and backend. After that I might add a few examples in swagger and then this would be ready to merge.

doc/decisions/backend_api_documentation.md
```diff
@@ -0,0 +1,52 @@
+# Backend API documentation
+
+## Problem
+
+We need a way to document the API in the backend. Otherwise we would have to read the Rust Code every time we want to develop a new part of the frontend.
+
+## Constraints
+
+## Assumptions
+
```
markus2330, line 10:
```suggestion

- the backend's API is only used by the frontend

```

doc/decisions/backend_api_documentation.md
```diff
@@ -0,0 +1,52 @@
...
+-   When deciding on a new endpoint it will take longer for the API documentation to be done, as the backend code has to be finished first.
+-   The API documentation will not deviate from the actual implementation as it is generated from code.
+-   Developers have to know Swagger, but once they do they can easily check the API in the browser served by the backend.
+-   Backend Developers have to invest some time into reading Utoipa's documentation.
+
+## Related Decisions
+
+## Notes
+
+Once the project is in a later stage we might consider additionally using API Blueprint. At that point a cleaner API becomes more important and making changes requires more careful consideration.
```
markus2330, line 52: Not needed to be written. If assumptions are broken, the decision needs to be reevaluated again.

```suggestion
```

doc/decisions/backend_api_documentation.md
```diff
@@ -0,0 +1,52 @@
...
+
+### API Blueprint
+
+Information from <https://testfully.io/blog/api-blueprint/>.
+
+Before starting on a new endpoint we would write a markdown document specifying in detail what the endpoint should return and in what format. There is also a [lot of tooling](https://apiblueprint.org/tools.html) for API Blueprint.
+
+## Decision
+
+Utoipa
```
markus2330, line 31: Write a bit how exactly we want to use it.

doc/decisions/backend_api_documentation.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Rationale
+
+With Utoipa it is possible to generate API documentation directly from code whereas with API Blueprint we would have to write the API documentation by hand in markdown files (at least I have yet to find a crate that automatically generates required markdown files). Additionally if we where to use API Blueprint every time the API changes we would have to remember to change the markdown files as well, which is easily forgotten and might become tedious when making smaller changes.
+
+For me (@kitzbergerg) it is easier to use Utoipa as it generates Swagger/OpenApi which I have already used. I would have to learn API Blueprint and its tooling from scratch. With that I would argue that more people know Swagger/OpenApi leading to easier onboarding.
+
+One advantage of API Blueprint would be that we have to carefully think about the API (API-First Approach) before actually implementing. This leads to a clean API in the long run, however it is more difficult to quickly prototype and try different approaches (which in my opinion is currently more important as we are somewhat at the beginning of the project).
+
```
markus2330, line 40: Please write about which data types are supported by Utoipa, e.g., are SQL or Diesel Types supported, e.g. https://lib.rs/crates/postgis_diesel? How much time would it need to add support for them?

doc/decisions/backend_api_documentation.md
```diff
@@ -0,0 +1,52 @@
...
+Compared to swagger in Java there is a bit more you have to configure to get it running.
+
+Once you get it running you get endpoints with schemas and examples representing the API. You can also execute example requests directly from the UI to see how the backend behaves.
+
+### API Blueprint
+
+Information from <https://testfully.io/blog/api-blueprint/>.
+
+Before starting on a new endpoint we would write a markdown document specifying in detail what the endpoint should return and in what format. There is also a [lot of tooling](https://apiblueprint.org/tools.html) for API Blueprint.
+
```
markus2330, line 28: Please write also about other alternatives you found. Some investigations should be done.

markus2330 (changes requested):
> @markus2330 I just pushed the decision. Is that what you are looking for?

Yes, very interesting proposal.

> If that is ok I would continue with Utoipa and as a next step remove the Body struct from frontend and backend. After that I might add a few examples in swagger and then this would be ready to merge.

You can invest a bit more time to present Utopia to our team better. It is not yet decided if we use it, people who need to write Utopia code first need to have a look at it.

backend/README.md
```diff
@@ -61,7 +61,9 @@ Example requests:
 
 ## Documentation
 
-To view the documentation run
+To view api documentation go to <http://localhost:8080/swagger-ui/>.
```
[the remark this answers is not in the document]
kitzbergerg: Those two endpoints are the ones used in every example from Utoipa. For Spring it is as far as I know somewhat common to have the endpoint like this so I think leaving it like that is a good option (although I can easily change them if another endpoint is preferred).

A bit of additional information: The `swagger-ui` endpoint actually serves content showing the API (official demo: https://petstore.swagger.io/). The `/api/openapi.json` is actually just quite unreadable json so I will probably never be used apart from swagger-ui.

backend/src/config/routes.rs
```diff
@@ -14,6 +23,49 @@ pub fn config(cfg: &mut web::ServiceConfig) {
...
+        tags(
+            (name = "plants")
+        )
+    )]
+    struct PlantsApiDoc;
+
+    let mut openapi = SeedApiDoc::openapi();
+    openapi.merge(PlantsApiDoc::openapi());
+
+    cfg.service(SwaggerUi::new("/swagger-ui/{_:.*}").url("/api-doc/openapi.json", openapi));
```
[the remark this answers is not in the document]
kitzbergerg: See https://github.com/ElektraInitiative/PermaplanT/pull/101#discussion_r1134586555.
The weird character apparently have special meaning in Actix. At least is was like this in the utoipa example with Actix (https://github.com/juhaku/utoipa/blob/master/examples/actix-web-multiple-api-docs-with-scopes/src/main.rs#L27)

backend/src/controller/plants.rs
```diff
@@ -12,6 +12,12 @@ use actix_web::{
 /// # Errors
 /// * If the connection to the database could not be established.
 /// * If [web::block] fails.
+#[utoipa::path(
+    context_path = "/api/plants",
```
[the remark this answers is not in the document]
kitzbergerg: I will look into it, but I don't think so. At least every example I found does it like this. I would guess that is quite difficult (if not impossible) for the macro Utoipa uses to automatically infer the actual path from the routes file that is somewhere else in the code.

backend/src/main.rs
```diff
@@ -77,9 +77,9 @@ async fn main() -> std::io::Result<()> {
 
         App::new()
             .wrap(cors_configuration())
-            .wrap(NormalizePath::default())
```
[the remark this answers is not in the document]
kitzbergerg: That seems to be a bug with NormalizePath in combination with utoipa. After trying and playing around for quite some time this was the only difference between the examples and my code and after removing it the swagger-ui worked. The other APIs are still normalized (I moved `NormalizePath` into the routes.rs file).

backend/src/model/response.rs
```diff
@@ -2,10 +2,19 @@
...
+use utoipa::ToSchema;
+
+use super::dto::{PlantsDto, SeedDto};
 
 /// The default return type of the server.
 #[typeshare]
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+#[aliases(
+    BodyVecSeed = Body<VecSeed>, VecSeed = Vec<SeedDto>, // TODO: not rendered as array in swagger
```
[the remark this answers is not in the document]
kitzbergerg: That is a problem with generics (it's one of the reasons why I want to remove the `Body` struct).
So no what is currently rendered in swagger-ui is actually not entirely correct (that's why the TODO).
Maybe there is some other way to do this, but utoipa is somewhat new and there are no examples on using nested generics. After some digging the best I found was https://docs.rs/utoipa/3.1.0/utoipa/derive.ToSchema.html#generic-schemas-with-aliases.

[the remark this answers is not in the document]
markus2330: Please report this problem upstream. If it is a new project and has some obvious bugs it is imho okay, if they at least fix such problems quickly. It is only by chance that in this case we could remove the `Body`  struct. In other cases, this might not be possible.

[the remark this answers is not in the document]
kitzbergerg: See https://github.com/juhaku/utoipa/issues/524

backend/src/config/routes.rs
```diff
@@ -14,6 +23,49 @@ pub fn config(cfg: &mut web::ServiceConfig) {
...
             )
-            .service(web::scope("/plants").service(plants::find_all)),
+            .service(web::scope("/plants").service(plants::find_all))
+            .wrap(NormalizePath::default()),
     );
 }
+
+/// Creates a swagger doc endpoint
+pub fn api_doc(cfg: &mut web::ServiceConfig) {
+    /// Struct used by utoipa to generate seeds-api-doc
```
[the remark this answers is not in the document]
kitzbergerg: I added a bit more information. But I would leave it at that, for more information developers should read the Utoipa docs. At least in my opinion it is not the function documentations job to tell how to extend the function but rather how to use it.

If we really want additional information about how to implement a new endpoint with Utoipa I would make that part of the mdBook.

doc/decisions/backend_api_documentation.md
```diff
@@ -0,0 +1,52 @@
...
+For me (@kitzbergerg) it is easier to use Utoipa as it generates Swagger/OpenApi which I have already used. I would have to learn API Blueprint and its tooling from scratch. With that I would argue that more people know Swagger/OpenApi leading to easier onboarding.
+
+One advantage of API Blueprint would be that we have to carefully think about the API (API-First Approach) before actually implementing. This leads to a clean API in the long run, however it is more difficult to quickly prototype and try different approaches (which in my opinion is currently more important as we are somewhat at the beginning of the project).
+
+## Implications
+
+-   When deciding on a new endpoint it will take longer for the API documentation to be done, as the backend code has to be finished first.
```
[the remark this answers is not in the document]
kitzbergerg: What do you mean by that?
You have to derive ToSchema and you also might need to add something to the API config. But do we want implementation details like that in the decision?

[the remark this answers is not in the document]
markus2330: In this case it seem to be relevant . At least I would have understood with "backend code has to be finished" nearly the opposite to "You have to derive ToSchema and you also might need to add something to the API config".

```suggestion
-   When deciding on a new endpoint it will take longer for the API documentation to be done, as:
    - the Rust structs must be implemented
    - `derive ToSchema` must be present for all structs, and
    - you also might need to change the API config.
```

backend/src/controller/plants.rs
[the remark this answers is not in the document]
kitzbergerg: I have found nothing that would suggest a way to remove it. Maybe when Utoipa improves in the future, but like I mentioned it seems quite difficult to do that in Rust.

doc/decisions/backend_api_documentation.md
kitzbergerg: I don't quite understand this question. As far as I can tell all basic Rust types are supported as well as some a bit more special ones: https://github.com/juhaku/utoipa#crate-features. We also only have to use Utoipa for the DTOs, the types used in the database should not be exposed in the API anyways.

Apart from that there is also the option to define `#[schema(value_type = String, format = DateTime)]` so that another type or some format is used in the API doc.

doc/decisions/backend_api_documentation.md
kitzbergerg: Do you mean alternatives in the sense of crates/mechanisms similar to Utoipa or API blueprint? Or rather investigate API blueprint in more detail?

markus2330:
I removed this from the project (we only add issues to the project)

backend/src/controller/seed.rs
```diff
@@ -6,41 +6,56 @@ use actix_web::{
...
 /// # Errors
 /// * If the connection to the database could not be established.
 /// * If [web::block] fails.
+#[utoipa::path(
+    context_path = "/api/seeds",
+    responses(
+        (status = 200, description = "Fetch all seeds", body = Vec<SeedDto>)
+    )
+)]
 #[get("")]
```
Benjaminimal, line 22: I think we can omit `context_path` if we omit the route in `config::routes` and set it here in `#[get("/api/seeds")]`.
Not sure if that's the best way to do that.
I'll look into the whole PR later in the evening.

kitzbergerg: Yes that would be possible. If you think that is easier to understand then I can change it.

Benjaminimal: I think it would be easier to maintain since we don't have to repeat the path in two different files that way.

kitzbergerg: Hm. Seems to be more annoying than I thought. There is no real way to specify an empty scope/empty service in actix. So I cannot group routes together, which leads to problems with the API documentation and NormalizePath.
It would technically be possible to do, but is kind of hacky, so I would prefer to leave it like it is for now.

Benjaminimal (approved):

markus2330 (approved):

doc/decisions/backend_api_documentation.md
markus2330: I meant other crates but you already said it in the meeting: it seems like there is not really something else available.

PR #131, opened 2023-03-20
117 nextcloud integration decision - authentication

absurd-turtle:
## Basics

## Checklist

- [ ] ~~I added unit tests for my code~~
      (not in the PR description)
- [ ] ~~I added code comments, logging, and assertions as appropriate~~
- [x] I mentioned [every code](/.reuse/dep5) not directly written by me in [reuse syntax](https://reuse.software/)

## Review

### DIFF
--- a/doc/decisions/auth.md
+++ b/doc/decisions/auth.md
@@ -77,7 +77,11 @@ Client 2: PermaplanT
 **Notes:**
 In this approach the accounts are managed by PermaplanT,
 therefore requires implementing an identity provider in PermaplanT.
-Alternatively to developing an identity provider a service like [keycloak](https://www.keycloak.org/) can be used.
+
+## Decision
+
+We will use [Keycloak](https://www.keycloak.org/).
+
 
 ## Further Notes
 There was an attempt to implement OIDC in nextcloud/server but the issue lost traction and was finally not merged because of missing integration tests (https://github.com/nextcloud/server/pull/12567)
--- a/doc/decisions/auth.md
+++ b/doc/decisions/auth.md
@@ -76,7 +76,7 @@ Client 2: PermaplanT
 **Notes:**
 In this approach the accounts are managed by PermaplanT,
 therefore requires implementing an identity provider in PermaplanT.
-Alternatively to developing an identity provider a servive like [keycloak](https://www.keycloak.org/) can be used.
+Alternatively to developing an identity provider a service like [keycloak](https://www.keycloak.org/) can be used.
 
 ## Further Notes
 There was an attempt to implement OIDC in nextcloud/server but the issue lost traction and was finally not merged because of missing integration tests (https://github.com/nextcloud/server/pull/12567)
--- a/doc/decisions/auth.md
+++ b/doc/decisions/auth.md
@@ -56,7 +56,7 @@ OpenID Connect authentication & authorization flow with:
 Authorization server: PermaplanT
 Resource server 1: Nextcloud
 Resource server 2: PermaplanT
-Client 1: [Nextcloud WebClient with oidc login](https://github.com/pulsejet/nextcloud-oidc-login)
+Client 1: Nextcloud WebClient [with oidc login](https://github.com/pulsejet/nextcloud-oidc-login) or [user\_oidc](https://github.com/nextcloud/user_oidc)
 Client 2: PermaplanT
 
 **Notes:**
@@ -64,3 +64,5 @@ In this approach the accounts are managed by PermaplanT,
 therefore requires implementing an identity provider in PermaplanT.
 Alternatively to developing an identity provider a servive like [keycloak](https://www.keycloak.org/) can be used.
 
+## Further Notes
+There was an attempt to implement OIDC in nextcloud/server but the issue lost traction and was finally not merged because of missing integration tests (https://github.com/nextcloud/server/pull/12567)
--- a/doc/decisions/auth.md
+++ b/doc/decisions/auth.md
@@ -23,7 +23,8 @@ Then they get the appropriate authorization (does the user have the right privil
 
 ## Solutions
 
-### Alternative A
+### Nextcloud-only
+
 OpenID Connect authentication & authorization flow with:
 
 Authorization server: [Nextcloud OIDC App](https://github.com/H2CK/oidc)
--- a/doc/decisions/auth.md
+++ b/doc/decisions/auth.md
@@ -3,9 +3,8 @@
 ## Problem
 The user wants to log in once and be able to fetch resources
 from the Nextcloud instance as well as from PermaplanT database.
-The user needs to authenticate themself (who is the user)
-and get the appropriate authorization (does the user have the right priviliges)
-to perform the action.
+The users need to authenticate themself (who is the user).
+Then they get the appropriate authorization (does the user have the right privileges) to perform the action.
 
 **OAuth2.0:** **User** can authorize **client** to fetch resources from **resource server**.
 
--- a/doc/decisions/auth.md
+++ b/doc/decisions/auth.md
@@ -1,8 +1,8 @@
 # Authentication & Authorization
 
 ## Problem
-The user wants to log in once and be able to fetch ressources
-from the Nextcloud Instance as well as from PermaplanT database.
+The user wants to log in once and be able to fetch resources
+from the Nextcloud instance as well as from PermaplanT database.
 The user needs to authenticate themself (who is the user)
 and get the appropriate authorization (does the user have the right priviliges)
 to perform the action.
--- a/doc/decisions/auth.md
+++ b/doc/decisions/auth.md
@@ -93,6 +93,18 @@ The only limitation is that Nextcloud does not offer scoped access at the moment
 
 We will use oidc with the separate identity provider [Keycloak](https://www.keycloak.org/).
 
+Keycloak itself will run as a separate service like Nextcloud. Users are created there instead of in Nextcloud/PermaplanT and users can then log in via Keycloak.
+
+So in the PermaplanT frontend the user will be redirected to Keycloak. They can login in Keycloak and then they get redirected back to PermaplanT.
+From then on all requests are authorized via JWT.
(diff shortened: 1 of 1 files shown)


absurd-turtle:
I investigated different ways to authenticate PermaplanT with Nextcloud.
I haven't made up my mind what the best of the proposed solutions is. 
All of them have their own issues and/or uncertainties.

kitzbergerg:
First of all I would strongly discourage anything that has to do with handling users or authentication with custom code (e.g. in the backend). There are way to many things to worry about. We should leave that to people who know (browser) security and simply use an authentication server and oauth2 libraries.

I think the decision currently mainly depends on two points:

1. How large do we think the backend will actually be in the end? If there are only a few endpoints it doesn't matter if Nextcloud doesn't support scopes.
2. Are there other applications apart from the PermaplanT backend and Nextcloud that might at some point use authentication? In that case it is probably worth it to use Keycloak.

I have used Keycloak already and it is quite nice. There are however a few problems with the Keycloak approach:
1. It is a separate application requiring a separate database (e.g. postgres). So setup might be complicated.
2. It is probably very difficult for students that have never used Keycloak to manage it. OAuth2 and Keycloak are quite complicated. So Keycloak is something @markus2330 would have to manage in the long run.

Btw if you need help with OAuth2/Keycloak I know a lot about that ;)

doc/decisions/auth.md
```diff
@@ -0,0 +1,76 @@
+# Authentication & Authorization
+
+## Problem
+The user wants to log in once and be able to fetch ressources
+from the Nextcloud Instance as well as from PermaplanT database.
```
markus2330, line 5:
```suggestion
The user wants to log in once and be able to fetch resources
from the Nextcloud instance as well as from PermaplanT database.
```

doc/decisions/auth.md
```diff
@@ -0,0 +1,76 @@
...
+The user needs to authenticate themself (who is the user)
+and get the appropriate authorization (does the user have the right priviliges)
+to perform the action.
```
markus2330, line 8:
```suggestion
The users need to authenticate themself (who is the user).
Then they get the appropriate authorization (does the user have the right privileges) to perform the action.
```

doc/decisions/auth.md
```diff
@@ -0,0 +1,76 @@
...
+Resource server 2: PermaplanT
+
+Client 1: Nextcloud WebClient [with oidc login](https://github.com/pulsejet/nextcloud-oidc-login) or [user\_oidc](https://github.com/nextcloud/user_oidc)
+
+Client 2: PermaplanT
+
+**Notes:**
+In this approach the accounts are managed by PermaplanT,
+therefore requires implementing an identity provider in PermaplanT.
+Alternatively to developing an identity provider a servive like [keycloak](https://www.keycloak.org/) can be used.
```
markus2330, line 73:
```suggestion
Alternatively to developing an identity provider a service like [keycloak](https://www.keycloak.org/) can be used.
```

doc/decisions/auth.md
```diff
@@ -0,0 +1,76 @@
...
+Resource server 1: [Nextcloud OAuth2](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/oauth2.html)
+
+Resource server 2: PermaplanT backend
+
+Client: PermaplanT frontend
+
+**Notes:**
+> Currently it is not yet possible to use an issued Access Token or ID Token to access resources at the Nextcloud instance it self.
+
+\- [H2CK/oidc](https://github.com/H2CK/oidc)
```
markus2330, line 34: I think something went wrong here. Did you want to add the link with the bug? Can we maybe fix this problem?

doc/decisions/auth.md
```diff
@@ -0,0 +1,76 @@
...
+to perform the action.
+
+**OAuth2.0:** **User** can authorize **client** to fetch resources from **resource server**.
+
+**OpenID Connect:** Extension of OAuth2.0 for authentication
+
+## Constraints
+
+- single account to access PermaplanT and Nextcloud
+
```
markus2330, line 17:
```suggestion
- we don't implement our own auth solution

## Assumptions

- we don't need roles (everyone can access all parts of the API)
- access control is quite simple, e.g., which users can read or write on which map

```

markus2330 (changes requested):
Thank you for starting this! I am not sure if I understand all alternatives and their pro/cons.

markus2330:
> First of all I would strongly discourage anything that has to do with handling users or authentication with custom code (e.g. in the backend). There are way to many things to worry about. We should leave that to people who know (browser) security and simply use an authentication server and oauth2 libraries.

Fully agree, I suggested the constraint. @kitzbergerg please also feel free to add such suggestions, this makes it easier for the person doing the decision. You can even push to the branch if you are sure about something.

> I think the decision currently mainly depends on two points:

>  How large do we think the backend will actually be in the end? If there are only a few endpoints it doesn't matter if Nextcloud doesn't support scopes.

I suggested this as an assumption.

>   Are there other applications apart from the PermaplanT backend and Nextcloud that might at some point use authentication? In that case it is probably worth it to use Keycloak.

Nextcloud+PermaplanT backend does everything what our Verein wants to do. If others want to run even more services (E-Mail etc.) is of course unknown. I suggested an assumption about that, too.

> I have used Keycloak already and it is quite nice. There are however a few problems with the Keycloak approach:

>    It is a separate application requiring a separate database (e.g. postgres). So setup might be complicated.
>    It is probably very difficult for students that have never used Keycloak to manage it. OAuth2 and Keycloak are quite complicated. So Keycloak is something @markus2330 would have to manage in the long run.

Let us add this text to the decision. For me it is quite clear that I would prefer a Nextcloud-only variant.

> Btw if you need help with OAuth2/Keycloak I know a lot about that ;)

:heart:

doc/decisions/auth.md
absurd-turtle: I linked the github repository where this is stated under "Limitations". I can also add https://github.com/H2CK/oidc/issues/87 where this is discussed further.

doc/decisions/auth.md
```diff
@@ -0,0 +1,82 @@
...
+Resource server 2: PermaplanT
+
+Client 1: Nextcloud WebClient [with oidc login](https://github.com/pulsejet/nextcloud-oidc-login) or [user\_oidc](https://github.com/nextcloud/user_oidc)
+
+Client 2: PermaplanT
+
+**Notes:**
+In this approach the accounts are managed by PermaplanT,
+therefore requires implementing an identity provider in PermaplanT.
+Alternatively to developing an identity provider a service like [keycloak](https://www.keycloak.org/) can be used.
```
markus2330, line 79:
```suggestion

## Decision

We will use [Keycloak](https://www.keycloak.org/).

```

@kitzbergerg can you give some info about how we would use Keycloak? (Do we need adapters, how to configure it etc. pp.)

markus2330 (changes requested):
Please give the alternatives names. We will most likely go for the Keycloak solution, as everything else doesn't seem to work, so you can already add this as decision.

markus2330:
There are a few alternatives to Keycloak that should be added to the decision, too:

- https://gist.github.com/bmaupin/6878fae9abcb63ef43f8ac9b9de8fafd
- https://lemonldap-ng.org/

They only should be listed as reference, Keycloak seems to be best of them, especially considering @kitzbergerg has experience with it and @absurd-turtle already checked that it works.

@kitzbergerg do you also have experience with allowing login via other parties, e.g., GitHub?

doc/decisions/auth.md
kitzbergerg: For our usecases I don't think we need any custom code/logic. 
Keycloak itself would run as a separate service like Nextcloud. You create users there instead of in Nextcloud/PermaplanT and users can then log in via Keycloak. 

So in the PermaplanT frontend you would get a redirect to Keycloak, you login in Keycloak and then you get redirected back to PermaplanT. From then on all requests are authorized via JWT. How that happens in detail is specified by the OAuth2 spec and there are also multiple ways of doing that (I don't think we need to worry about that in detail just use the recommended variant which is [Authorization Code Flow with PKCE](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-proof-key-for-code-exchange-pkce)).
The backend you then simply validate the tokens and extract roles/user information out of them. There are probably libraries for that (although doing that part by hand is not that complicated).

Keycloak itself can either run as an executable or as a docker container. I have used it as a docker container (which would also be my preferred solution as configuration is easier that way). Apart from that it only requires a database so we can use our existing PostgreSQL for that.

kitzbergerg:
> @kitzbergerg do you also have experience with allowing login via other parties, e.g., GitHub?

You can configure Keycloak to also allow login via e.g. Github so that is included. How that works in detail I don't know. You might have to 'ask' for an OAuth2 client from Github. Login via Google, Facebook, ... is also supported.

doc/decisions/auth.md
markus2330: Thx, good suggestions, @absurd-turtle can you add this to the decision. (Except of Docker, we will probably use Ansible for deployment but this is not really relevant for this decision here.)

absurd-turtle:
> There are a few alternatives to Keycloak that should be added to the decision, too:
> 
>     * https://gist.github.com/bmaupin/6878fae9abcb63ef43f8ac9b9de8fafd
> 
>     * https://lemonldap-ng.org/
> 
> 
> They only should be listed as reference, Keycloak seems to be best of them, especially considering @kitzbergerg has experience with it and @absurd-turtle already checked that it works.
> 
> @kitzbergerg do you also have experience with allowing login via other parties, e.g., GitHub?

At the moment GitHub has OAuth capabilities but is not a fully-fledged Identity provider like Google or Facebook. [source](https://stackoverflow.com/questions/71741596/how-do-i-implement-social-login-with-github-accounts)

markus2330 (approved):
Great job! Can you add OAuth2.0, OpenID Connect etc. to our common glossary?

PR #145, opened 2023-03-21
load baseApiUrl from env file

absurd-turtle:
## Basics

## Checklist

      (not in the PR description)

## Review


frontend/src/config/index.ts
```diff
@@ -1 +1 @@
-export const baseApiUrl = 'http://localhost:8080';
+export const baseApiUrl = import.meta.env.VITE_BASE_API_URL;
```
Bushuo, line 1: Please add for better debugging experience.
```suggestion
if (!import.meta.env.VITE_BASE_API_URL) {
  throw 'VITE_BASE_API_URL no set';
}

export const baseApiUrl = import.meta.env.VITE_BASE_API_URL;
```

Or we could use a library like tiny-invariant. Makes for prettier code I think.
```ts
import invariant from 'tiny-invariant';

invariant(!import.meta.env.VITE_BASE_API_URL, 'VITE_BASE_API_URL no set');
```

https://www.npmjs.com/package/tiny-invariant

Bushuo (commented):
Nice 💯

frontend/src/config/index.ts
```diff
@@ -1 +1,5 @@
-export const baseApiUrl = 'http://localhost:8080';
+if (!import.meta.env.VITE_BASE_API_URL) {
+  throw 'VITE_BASE_API_URL no set';
```
markus2330, line 2:
```suggestion
  throw 'VITE_BASE_API_URL not set';
```

frontend/sample_env
```diff
@@ -0,0 +1,4 @@
+// Env variables must have the prefix VITE_ for Vite to include them.
```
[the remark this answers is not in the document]
markus2330: I agree!

markus2330 (approved):
Thank you! Other than renames and typo it looks good to me (didn't try if it works, though).

frontend/src/config/index.ts
markus2330: Better be careful with (tiny) libs/packages. If really needed we should give this a decision, maybe something like https://github.com/jaredpalmer/tsdx#invariant is better suited?

Benjaminimal: I agree with not throwing too many packages at the project.
However, I would go with the defensive route of checking that this is set.
Not only is it good for developer experience but also when deploying to make sure this is set.
However, throwing a string literal seems like bad practice since catching it seems messy.
Maybe throw a subclass of `Error`, e.g. `ConfigurationError`?

Bushuo: >However, throwing a string literal seems like bad practice since catching it would be messy.
Maybe throw a subclass of Error, e.g. ConfigurationError?

The way it is setup currently, we would not catch errors from the config, because they happen before the app renders for the first time. But I agree, if we would ever want to, throwing an `Error` would be more sane, sorry for that 😅

Bushuo (approved):

markus2330:
jenkins build please

markus2330:
jenkins build please

absurd-turtle:
The .env file still has to be created in the deployment process. @lukashartl How should this be done?

lukashartl:
> The .env file still has to be created in the deployment process. @lukashartl How should this be done?

I'll take care of it :)

frontend/src/config/index.ts
lukashartl: I think the suggestion is better than adding a tiny lib just for a single function call. Reminds me a bit of the [leftpad story](https://www.davidhaney.io/npm-left-pad-have-we-forgotten-how-to-program/).

+1 for throwing an error instead of a string :)

lukashartl (approved):

lukashartl:
@absurd-turtle one question remains for me:

Is it possible to set `VITE_BASE_API_URL` to something relative like `/api`, so it automatically (or more like implicitly) uses `window.location.origin` as it's base?

Because if not, we would need to build separate frontends for each instance; only because the domains are different ...

markus2330:
> Because if not, we would need to build separate frontends for each instance; only because the domains are different ...

@lukashartl correct me if I am wrong but I don't think this is a problem, as we have 3 different nodes with 3 different deployment jobs.

lukashartl:
I'll hijack this branch now 😄 I'll rebase it, add the CI stuff and check if the pr instance works now 😄
If everything looks good, I'll also merge it right away.

lukashartl:
PR instance works now. Well done to everyone 🥇 

> @lukashartl correct me if I am wrong but I don't think this is a problem, as we have 3 different nodes with 3 different deployment jobs.

I had to change the logic, because an empty string would also evaluate to false. But in our case, if we set the env variable VITE_BASE_API_URL, but leave it empty, we don't have to pass in the domain name at all! Meaning we can really promote PR to Dev without having to fully recompile the whole frontend! So we actually save time in the pipeline 🕐 

@markus2330 pls have a look and if you are ok with my change, you can merge it right away :rocket: :rocket:

frontend/src/config/index.ts
lukashartl: I am using `import.meta.env.hasOwnProperty` now, so that `VITE_BASE_API_URL` can be an empty string without failing.

lukashartl (approved):

markus2330:
Great job, it is so beautiful to see such collaboration :revolving_hearts:

lukashartl:
Wonderful to see that PR and Dev are actually usable now (e.g. XHR is working now) 😀
And the new landing page as well 🥹

markus2330:
Yes, amazing, it is working wonderful now! :revolving_hearts:

Issue #98, opened 2023-03-12
CI frontend


markus2330:
CI features not related to deployment, to be executed as separate build job on our Docker agents:

- [ ] unit tests?
- [ ] reformatting?
- [ ] linter?
- [ ] import database in PR tests (scraper)

(? refers that we don't know how to do it yet, not that we don't want it)

markus2330:
I moved some tasks to #63, here only the frontend tasks remain.

@badnames @aidnurs  can you help out here a bit which commands are needed to do all tests, building of docu, reformatting, etc.?

@lukashartl All the jobs should be executed in parallel (backend+frontend).

aidnurs:
Hey guys,

> can you help out here a bit which commands are needed to do all tests, building of docu, reformatting, etc.?

I suppose `npm run build` executes the linter as well, right @buenaflor? I suggest to setup the linter so that it fails if the formatting doesn't correspond to our prettier configs in `prettier.config.js` instead of the pipeline formats every file automatically on the build itself, so the developer is aware of every change.

Additionally, we could add `npm run test` to `package.json` doing nothing for the time being, so we have the CD pipeline fixed and ready. 

Except for these 2 things, we should be good. If something additional regarding the topic comes to my mind, I'll let you know for sure.

markus2330:
Have we already enabled [all the eslint-plugin-react-hooks rules](https://react.dev/learn/editor-setup#linting)?

Is the code currently formatted via https://prettier.io?

buenaflor:
`npx lint-staged` should be the command if you take a look at `package.json` 

```
  "lint-staged": {
    "src/**/*.{js,jsx,ts,tsx,json,css,scss,md}": [
      "prettier --write"
    ],
    "src/**/*.{js,jsx,ts,tsx,json}": [
      "eslint --max-warnings=0"
    ]
  }

```

Bushuo:
@buenaflor
I think we need two more scripts for that. If nothing is staged in the commit, `npx lint-staged` won't do anything
```zsh
# paulb in ~/Projects/uni/bakk/PermaplanT/frontend on git:authors o [12:37:19] 
$ npx lint-staged
→ No staged files found.
```

Usually I have scripts like this in my projects
```json
// package.json
...
"scripts": {
  "format": "prettier --write .",
   "lint": "eslint --cache --cache-location ./node_modules/.cache/eslint ."
}
...
```

I did not try this yet. I can earliest in the evening today.

PR #147, opened 2023-03-21
fix(seeds): better error messages

Bushuo:
This PR adds an interceptor to the axios response in case of an error, and maps the message to a more user friendly string.

## Basics

- [X] Details of what you changed are in commit messages
- [X] The PR is rebased with current master.

## Checklist

      (not in the PR description)
- [X] I added code comments, logging, and assertions as appropriate

## Review

### DIFF
--- a/frontend/src/config/index.ts
+++ b/frontend/src/config/index.ts
@@ -1,1 +1,18 @@
+import mapErrorToString from '@/utils/map-error-to-string';
+import axios from 'axios';
+
+// Intercept the axios response to map errors to strings.
+axios.interceptors.response.use(
+  (r) => r,
+  (error) => {
+    console.error(error);
+
+    if (error instanceof Error) {
+      error.message = mapErrorToString(error);
+    }
+
+    return Promise.reject(error);
+  },
+);
+
 export const baseApiUrl = 'http://localhost:8080';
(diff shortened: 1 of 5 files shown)


frontend/src/config/index.ts
```diff
@@ -1 +1,18 @@
...
+
+// Intercept the axios response to map errors to strings.
+axios.interceptors.response.use(
+  (r) => r,
+  (error) => {
+    console.error(error);
+
+    if (error instanceof Error) {
+      error.message = mapErrorToString(error);
+    }
```
Benjaminimal, line 12: Do you think we really need this guard here?
What else could be passed to this error handler?
From the docs:
> Any status codes that falls outside the range of 2xx cause this function to trigger
https://axios-http.com/docs/interceptors

Benjaminimal (commented):
Looks good but someone should run `prettier` over the whole frontend sub dir.
Reviews with formatting changes are a pain.

markus2330:
> Looks good but someone should run prettier over the whole frontend sub dir.
> Reviews with formatting changes are a pain.

Absolutely, please propose in #98 how exactly we should do the reformatting (exact command, arguments, config).

The reformatting, however, should happen in an extra PR. Till then it is best to do as little reformatting as possible.

frontend/src/config/index.ts
Bushuo: Yes you're right :) I will remove the guard

Bushuo:
I agree with both of you. Thanks for the feedback 🎉🎉

PR #172, opened 2023-03-26
Add backend ORM decision resolves #170

### LINKED ISSUE #170: Decision: Which ORM/SQL crate do we want to use in the backend.
Currently we are using diesel as our ORM in the backend. There are other options such as `diesel_async`, `sea-orm` or simple sql crates like `sqlx`. Maybe one of them is a better fit for us.

kitzbergerg:
## Basics

## Checklist

      (not in the PR description)

## Review


kitzbergerg:
@buenaflor I added you as well as you were the one who originally implemented diesel. Maybe you want to have a look as well.

Benjaminimal:
There is one more thing we could note here.
Since we want to use postgis we also want to use an ORM which supports this extension.

doc/decisions/backend_orm_crate.md
```diff
@@ -0,0 +1,70 @@
+# Backend ORM/SQL crate
+
+## Problem
+
+We are currently using [diesel](https://github.com/diesel-rs/diesel) as our ORM in Rust.  
+Since diesel doesn't support async out of the box we will look at other alteratives and see if a switch is reasonable.  
+There might also be advantages of other ORM crates that as of writing this have not been considered.
```
markus2330, line 7: This is not a problem (unless it is clear why we would need async). The problem below could be used: that users might use async wrongly.

doc/decisions/backend_orm_crate.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Constraints
+
+1. Migrations need to be supported
+
+## Assumptions
+
+1. diesel is a solid choice, therefore unless other ORMs have obvious advantages differences will not be looked at in detail
```
markus2330, line 15: :+1:

doc/decisions/backend_orm_crate.md
```diff
@@ -0,0 +1,70 @@
...
+diesel_async is a drop-in replacement of corresponding diesel methods providing async support.
+
+### [sea-orm](https://github.com/SeaQL/sea-orm)
+
+Sea-orm is based upon [sqlx](https://github.com/jmoiron/sqlx) and supports async out of the box.  
+It supports [migrations](https://www.sea-ql.org/SeaORM/docs/migration/setting-up-migration/) written in Rust.
+
+## Decision
+
+Use diesel + diesel_async. There are no obvious disadvantages to diesel_async and required code changes are minimal.
```
markus2330, line 37: Please answer here: What did you check for? Is it as reliable, tested, ...?

doc/decisions/backend_orm_crate.md
```diff
@@ -0,0 +1,70 @@
...
+
+### async
+
+Both diesel_async and sea-orm provide async support.
+
+This might not necessarily lead to a performance increase, but it is easier for (especially inexperienced) developers to stay in full async.  
+
+For example if you forget `web::block` when executing a query using diesel you block the async executor (tokio in our case) until the query is finished.  
+We currently have a lint warning about async functions that do not use await, you will therefore notice this mistake.  
+However for inexperienced Rust developers it might be difficult to find out why this warning occurred as the code compiles and runs perfectly fine (especially with small loads).
```
markus2330, line 54: This could be used as problem. Did it actually occur already?

doc/decisions/backend_orm_crate.md
```diff
@@ -0,0 +1,70 @@
...
+
+### [diesel](https://github.com/diesel-rs/diesel) + [diesel_async](https://github.com/weiznich/diesel_async)
+
+diesel_async is a drop-in replacement of corresponding diesel methods providing async support.
+
+### [sea-orm](https://github.com/SeaQL/sea-orm)
+
+Sea-orm is based upon [sqlx](https://github.com/jmoiron/sqlx) and supports async out of the box.  
+It supports [migrations](https://www.sea-ql.org/SeaORM/docs/migration/setting-up-migration/) written in Rust.
+
```
markus2330, line 34: An alternative solution is that we review for correct async usage and you help if needed.

doc/decisions/backend_orm_crate.md
```diff
@@ -0,0 +1,70 @@
...
+This might not necessarily lead to a performance increase, but it is easier for (especially inexperienced) developers to stay in full async.  
+
+For example if you forget `web::block` when executing a query using diesel you block the async executor (tokio in our case) until the query is finished.  
+We currently have a lint warning about async functions that do not use await, you will therefore notice this mistake.  
+However for inexperienced Rust developers it might be difficult to find out why this warning occurred as the code compiles and runs perfectly fine (especially with small loads).
+
+### Performance 
+
+Metrics by diesel: <https://github.com/diesel-rs/metrics/>
```
markus2330, line 58: Please use normal markdown links.

markus2330 (changes requested):
Thank you, great job! For me it looks like we should simply learn async and review properly.

doc/decisions/backend_orm_crate.md
kitzbergerg: As far as I can remember it was created by the creator of diesel. I haven't looked at the actual implementation though. According to the examples the only difference to normal diesel is that you import different traits and in the performance metrics there is only a minimal decrease.
I would however say that it cannot be as tested as just diesel itself. It has way less downloads, but like I mentioned I believe it is easier to use async all the way through that having to worry about blocking functions.

markus2330: Please write this in the decision itself. Also if it is feature-complete etc.

markus2330 (changes requested):
Can you also add code examples which intermixing is problematic and should be avoided? If this is well described, then maybe simply reading the decision is enough to not fall into the trap?

Please write in the decision if you already found such wrong usages (and fixed it).

doc/decisions/backend_orm_crate.md
kitzbergerg: In my opinion that is not really an alternative solution, but would rather mean we just stick with diesel (without diesel_async). So I'm not adding a separate alternative.

Apart from that I'm not a fan of that as it would require manual checking every time and can be quite tricky to spot. For example currently simply calling any function in the `service` directory would lead to blocking of the executor. You cannot really spot blocking functions without actually knowing the whole codebase and which are blocking.
The same applies for locks and the likes. That is also why there are async lock such as [async RwLock](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html).

I added that to the decision as well.

kitzbergerg:
@markus2330 I update the decision. Unless we don't want to use diesel_async I would also implement the switch to diesel_async in this PR.

kitzbergerg:
@markus2330 This is ready to be merged. There is still a todo, but I would handle that as part of the testing strategy. I don't want to wait much longer otherwise there is merge mess again.

markus2330 (approved):

markus2330:
Thank you!

> I don't want to wait much longer

No need to say that, nobody wants to wait. Sometimes decisions are not so easy.

Issue #158, opened 2023-03-23
Rust mdBook


markus2330:
Intros:

- [x] add most important links (ideally also our API/code documentation)
- [x] it should be clear where which code should be added

Arc42:

- [x] integrate all current markdown files (architecture, usecases, etc.)
- [x]  quality goals:
        1. stability
        2. maintenance
        3. performance
        4. privacy
- [x] solution strategy
       - description of 3 layers
       - types/structs
       - diesel/rust intro in docs(Gabriel)
       - be stateless (all state in DB)

Issue #97, opened 2023-03-11
Add documentation for entites/DTOs in the backend


kitzbergerg:
As part of #60 all functions, structs and modules need to be documented. Apart from some structs representing the data this is done (#72). This issue represents those leftover structs that still need to be documented.

@markus2330 maybe you can find someone to do that (as I don't know what fields represent what data and how they are used).

kitzbergerg:
To find all structs that still need documentation search for `// TODO: See #97.`.

aidnurs:
Hey guys,

the single table, that I've defined in the database, is called `plant_detail`. The structs that require documentation in the backend were actually implemented by Giancarlo. Therefore, it would be a better match for Giancarlo to address this particular issue in my opinion.

aidnurs:
> @aidnurs it is about the plant_detail table (to be renamed to plants). The plants table is going to be removed (see #35).
> 
> If the structs don't exist already, it would be super-helpful if you can add them. If you struggle with the syntax, you can also add them commented-out.

Sure, I've updated the entity.rs with `plant_detail` and pushed it to #181, however, as I've said there are more structs and data tables marked to be documented, which I don't know much about.

markus2330:
@kitzbergerg is there something like `@copydoc` (from Doxygen) in Rust? Do you have some tips how we can document DTOs and Entities with no/little duplication?

kitzbergerg:
> Why would DTOs or Entities need that?

We don't need that for DTOs and Entities, only for functions. My comment was meant to give a general overview of I think we should use doc.

And like I mentioned I know of nothing/have found nothing similar to doxygen so far. 
Although at the same time I don't really think it is that necessary for Rust. 
It is also less to worry about when you change structs as you can just change the doc as well. 
As far as I'm concerned we don't have that many structs/things to document that we would need something like doxygen.

markus2330:
As discussed, will be done by @yvonnemarkl and @aidnurs

markus2330:
@aidnurs I unassigned you. Yvonne will do this after #229 is merged and you have finished your thesis.

Issue #39, opened 2023-02-03
testing strategy


markus2330:
Should be kicked-off as part of #158:

- [x] backend unit tests
- [x] frontend unit tests
- [x] backend integration tests
- [x] #38 
- [ ] manual tests (before releases etc.)
- [ ] UI component tests (storybook)

markus2330:
Do we need to split frontend testing strategy from this issue?

Bushuo:
> Do we need to split frontend testing strategy from this issue?

@markus2330 I don't think so. If we use Jest for testing it could be a benefit because of compatibility with Storybook. 
But it is not a must. We can go another route with the tests as well.
I do not really have that much experience with UI testing, and I just used Jest as a default.

badnames:
So, since frontend tests have been assigned to me, can we go ahead and add jest or do we still need a decision about that?

markus2330:
As you want it, I assigned you here for now.

kitzbergerg:
After looking through the code a bit I don't think we need unit tests currently. There isn't really any complicated functionality in any specific function that would require such in depth testing.

Integration tests however are very much necessary. According the [rust book](https://doc.rust-lang.org/book/ch11-03-test-organization.html) you should write integration tests entirely separate from your actual implementation. 
This would however mean we would have to split our current code into a library part and a binary part. The main function (binary part) would then only contain very simple stuff like reading config and then calling a function in the library that actually starts the server, while the library contains basically everything else.

I'll look around a bit to see how this is handled in other crates.

In the end I think for the backend we need a decision regarding that.
Separate from that I'll write a doc article where I explain when unit tests are necessary and how to properly set them up (although like I said we currently don't have/probably don't need unit tests).

kitzbergerg:
If you want to have a look at what splitting into binary and library in its simplest for looks like I did that for the mdbook preprocessor: https://github.com/kitzbergerg/mdbook-generate-summary/tree/main/src. If you look at the [main.rs](https://github.com/kitzbergerg/mdbook-generate-summary/blob/main/src/main.rs) there is actually no module defined. Instead the functions are imported from the `lib.rs` file.

markus2330:
Please write the testing strategy in a document in the repository, at least the things that are clear or we already do. It is better to continue discussion from a point where the basics/terminology are made clear first.

[...]

PR #198, opened 2023-03-31
Merge tables plants and plant_detail and make its data available in frontend.

badnames:
This pull request makes the scraped data by @aidnurs available in the "Add seeds" dialog.
For this purpose, plant_detail now replaces the old plants table.
A new variety / species search was added by @Bushuo.

## Basics

## Checklist

- [x] I added unit tests for my code
      (not in the PR description)
- [x] I mentioned [every code](/.reuse/dep5) not directly written by me in [reuse syntax](https://reuse.software/)

## Review


markus2330:
Great to see this PR! Seems like it doesn't build at the moment.

> This pull request makes the scraped data by @aidnurs available in the "Add plants" dialog.

Do you mean "Add seeds" here? Adding new plants is not part of this milestone.

backend/src/model/entity/plants_impl.rs
```diff
@@ -14,16 +16,43 @@ impl Plants {
...
+        let query_with_placeholders = format!("%{capitalized_query}%");
+
+        let query = plants::table
+            .select(all_columns)
+            .filter(
+                sql::<Bool>("LOWER(binomial_name) LIKE ")
+                    .bind::<Text, _>(&query_with_placeholders)
+                    .sql(" OR LOWER(ARRAY_TO_STRING(common_name, ' ')) LIKE ")
+                    .bind::<Text, _>(&query_with_placeholders),
+            )
```
Benjaminimal, line 43: If we have a powerful ORM we should use its features.
Only if we can't express our statement or they would be inefficient should we fall back to raw SQL.

There is no ned for `to_lowercase`, use ilike instead of like.
For filtering on the binomial name you can do this `.filter(binomail_name.ilike(query_with_placeholders))`.
https://docs.diesel.rs/master/diesel/expression_methods/trait.PgTextExpressionMethods.html#method.ilike

I'm not sure how to achieve something similar for the values in `common_name` but it can probably be done with the querybuilder as well.

backend/migrations/2023-03-29-111855_merge_plant_tables/up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE seeds DROP COLUMN plant_id;
+
+DROP TABLE plants;
+
+ALTER TABLE plant_detail RENAME TO plants;
+ALTER TABLE seeds ADD COLUMN plant_id INTEGER REFERENCES plants(id);
+
+INSERT INTO plants (id, binomial_name, common_name) VALUES (-1, 'Testia testia', ARRAY['Testplant']);
```
Benjaminimal, line 8: Test data shouldn't be part of migrations.
Please set up the test data in the test where you need it.
You can use this function to make sure nothing will be committed during a test to ensure test isolation.
https://docs.diesel.rs/master/diesel/connection/trait.Connection.html#method.test_transaction

A quick search led me to this example of how to use it.
https://github.com/diesel-rs/diesel/issues/1549#issuecomment-621792203

Benjaminimal (commented):
Great Job!
This should give us a clean start for all the further use cases.

backend/src/model/entity/plants_impl.rs
badnames: I opted to use raw SQL strings because I could not get the "correct" diesel ORM-methods to work.
Tbh. I can't quite remember what the exact problem was, but there was a trait missing somewhere. 

That being sain, I agree with your points and would prefer these changes if someone could get them to work.

backend/migrations/2023-03-29-111855_merge_plant_tables/up.sql
badnames: How are you finding these things when I've been sitting here googling this stuff for hours :sweat_smile:....

Anyways, thanks for the suggestion :+1: .

Benjaminimal: :shrug: 
Here is another example that is closer to our set up.
https://cloudmaker.dev/actix-integration-tests/

But maybe we can leave it for now and create a new issue to set up backend tests in a usable way?

backend/src/model/entity/plants_impl.rs
Bushuo: Did it :)

frontend/src/features/seeds/routes/CreateSeed.tsx
```diff
@@ -2,20 +2,48 @@ import PageLayout from '../../../components/Layout/PageLayout';
...
+
+    const timeout = setTimeout(() => {
+      searchFunction(searchParam);
+    }, delay);
+
+    setTimeoutId(timeout);
+  };
+
+  return [debouncedSearch];
+}
```
buenaflor, line 28: Debouncing might be useful for other components, maybe put this in the global `utils` folder

buenaflor (changes requested):

frontend/src/features/seeds/routes/CreateSeed.tsx
Bushuo: I agree.
I also changed the hook, so that it can be easily integrated into react-query, once we merge this.

backend/migrations/2023-03-29-111855_merge_plant_tables/up.sql
markus2330: Did you already see the tests we already have in backend/src/test/seed.rs?

Btw. @Benjaminimal **excellent** review.

markus2330: I agree there should be further tests, within M1.

backend/src/controller/plants.rs
```diff
@@ -1,25 +1,49 @@
...
-pub async fn find_all(pool: Data<Pool>) -> Result<HttpResponse> {
+pub async fn find_all_or_search(
+    query: Option<web::Query<PlantsSearchParameters>>,
+    pool: Data<Pool>,
+) -> Result<HttpResponse> {
+    if let Some(query) = query {
+        search(query, pool).await
+    } else {
+        find_all(pool).await
+    }
```
Benjaminimal, line 32: It might be cleaner if we don't call controllers from another controller.
You can directly go to the service layer like this.
```suggestion
    let response = match query {
        Some(parameters) => {
            web::block(move || service::plants::search(&pool, &parameters)).await??
        }
        None => web::block(move || service::plants::find_all(&pool)).await??,
    };
    Ok(HttpResponse::Ok().json(response))
```
You can also get rid of the other controllers.

backend/migrations/2023-03-29-111855_merge_plant_tables/up.sql
Benjaminimal: The seeds tests do the same.
Adding test data in migrations.
https://github.com/ElektraInitiative/PermaplanT/blob/9ac1b16809e94aa9cc6c5d8fadf41f8f74d4c249/backend/migrations/2023-01-21-024920_create_seeds/up.sql#L19-L22

No test transactions in the seeds test.
https://github.com/ElektraInitiative/PermaplanT/blob/9ac1b16809e94aa9cc6c5d8fadf41f8f74d4c249/backend/src/test/seed.rs

We also have a lot of code duplication here at the start of each test function.
This should be pulled out into some setup which also sets up a test database (maybe with some fixtures) so we can use test transactions easily.
We would also need a tear-down to drop this database after the last test has run.
I looked into it briefly and it seems like rust's built-in test crate doesn't provide easy setup and tear-down out of the box.
@kitzbergerg 
Do you know any recipe or good test crate to achieve this?

Bushuo: I added a test utility that initializes the pool so that it uses the test transactions.
This should be enough to cover easy test cases.
I made a comment about its limitations which @badnames and I identified.

kitzbergerg: As far as I know the tests in Rust shouldn't actually be used how we use them. They should be more on the side of unit tests rather than integration tests.
I just found a [crate](https://docs.rs/mockall/latest/mockall/) to mock traits (~interfaces). I think we should think about tests more carefully soon (especially as we currently require a running db that is not part of the backend) and split into unit and integration tests.

kitzbergerg:
That's a bit more general, but in my opinion this PR is way to large. With 51 changed files there is way to much to think about and it is quite difficult to read/see if there are any errors anywhere.

I haven't looked at the code in detail though, maybe there is a good reason/it is only possible in a large PR like this. 

Generally I would however prefer if PR wouldn't get that large. Maybe split the tasks into smaller subtasks next time. That way it is easier to write and review and you don't have huge merge conflicts with other branches before/after merging.

backend/src/model/dto.rs
```diff
@@ -53,11 +53,17 @@ pub struct NewSeedDto {
 
 #[allow(clippy::missing_docs_in_private_items)] // TODO: See #97.
```
kitzbergerg, line 54: Please remove lines `#[allow(clippy::missing_docs_in_private_items)] // TODO: See #97.` completely for all newly created enums/structs.
Instead you should document the structs and fields with code comments e.g. 
```rust
/// This represents ... and is used for ...
#[typeshare]
#[derive(Debug, Serialize, PartialEq, Eq, Deserialize, ToSchema)]
pub struct PlantsSummaryDto {
    /// The id of the entity
    pub id: i32,
   /// <Explaination of the field>
    pub binomial_name: String,
   /// <Explaination of the field>
    pub common_name: Option<Vec<Option<String>>>,
}
```

Please write comments that can be used to fully understand what the structs/fields do and what might be special about them. Those comments will later on be part of the documentation (generated from code).

You can look at the [doc of Vec](https://doc.rust-lang.org/std/vec/struct.Vec.html) and the [corresponding code](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#151) as and example. The doc is generated from code.

badnames:
> That's a bit more general, but in my opinion this PR is way to large. With 51 changed files there is way to much to think about and it is quite difficult to read/see if there are any errors anywhere.
> 
> I haven't looked at the code in detail though, maybe there is a good reason/it is only possible in a large PR like this.
> 
> Generally I would however prefer it if PRs don't get that large. Maybe split the tasks into smaller subtasks next time. That way it is easier to write and review and you don't have huge merge conflicts with other branches before/after merging.

Yes I agree.

The problem was that the goal posts moved quite a bit over time. At first we only wanted to add pagination for the existing plants-typeahead (hence the branch name). Then we discovered that the old plants table is no longer needed.

Since this is my first time working with actix-web and diesel I greatly underestimated the amount of work needed to merge these two tables.
While adding tests, we discovered that some refactoring was needed in out unit testing system (thanks again to @Benjaminimal).
And, since master was quite far ahead of our branch by this point, a lot of files needed to be changed to fix all merge conflicts.

So to summarize, let us all agree to only address one issue per pull request from now on.
I will therefore exclude the changes to our schema requested by @Benjaminimal and add them as a separate issue (again many thanks for pointing this out).

backend/migrations/2023-03-29-111855_merge_plant_tables/up.sql
markus2330: Yes, excellent suggestion, we definitely should separate unit and integration tests. There is #39, can you please kick this off as part of #158?

markus2330:
jenkins build please

buenaflor (approved):
👍

markus2330 (approved):
Amazing job! Great to finally see a search for plants. :1st_place_medal:

PR #222, opened 2023-04-04
Use Case: Watering layer

NurramoX:
This should document the following watering layer use case from #1 that was assigned to me

## Basics

- [x] Details of what you changed are in commit messages
      (first line should have `module: short statement` syntax)
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildserver is happy.
- [x] The PR is rebased with current master.

## Checklist

- [ ] I added unit tests for my code
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I fixed all affected decisions
- [ ] I added code comments, logging, and assertions as appropriate
- [ ] I mentioned [every code or binary](/.reuse/dep5) not directly written or done by me in [reuse syntax](https://reuse.software/)

## Review

- [ ] Documentation is conforming to [our Documentation Guidelines](/doc/documentation.md)
- [ ] Examples are well chosen and understandable
- [ ] Code is conforming to our Coding Guidelines
- [ ] Code is consistent to [our Design Decisions](/doc/decisions)

Issue #155, opened 2023-03-22
Revise database guidelines


NurramoX:
Database table names should be only plural to avoid inconsistencies

markus2330:
Thank you for creating the issue! Please check if there are already table names which need to be changed.

markus2330:
@NurramoX please also self-assign you to such issues, add it to the PermaplanT project and move it to the right place.

Issue #140, opened 2023-03-21
todo layer use case


absurd-turtle:

markus2330:
The last two use cases!

I added them to #1, to keep track when we finished all use cases (next week everything should be reviewed+merged, so it would be great if you can already write it this week.).

As written in #1:

- todo layer: create deck cards in your map
- offline availability (however, maybe notes can also be taken via label layer OR via drawing layer..?)

(@Bushuo this is one more use case that should work offline.)

PR #245, opened 2023-04-06
Add maps, users and blossoms DB schema and ER diagram

ThorbenStaufer:
Added table descriptions for maps, users and blossoms tables and updated the ER diagram to include them.

## Basics

## Checklist

      (not in the PR description)

## Review

### DIFF
--- a/doc/architecture/database_schemata.md
+++ b/doc/architecture/database_schemata.md
@@ -120,14 +120,26 @@ blossoms {
   VARCHAR title
   VARCHAR description
   VARCHAR condition
+  TRACKS track
   BYTEA icon
   BOOLEAN is_seasonal
 }
 
-tracks {}
+enum_tracks {
+  VARCHAR Beginners_Track
+  VARCHAR Seasonal_Track
+  VARCHAR Completionist_Track
+  VARCHAR Expert_Track
+}
 
-maps }o--|| users : ownes
-blossoms }o--|| tracks : ""
+blossoms_gained {
+  INT id PK
+  INT times_gained
+}
+
+maps }o--|| users : owned by
+blossoms ||--o{ blossoms_gained : ""
+blossoms_gained }o--|| users : ""
 
 ```
 
@@ -232,9 +244,16 @@ blossoms }o--|| tracks : ""
 | **title**                        | Novice Gardener                  |
 | **description**                  | Plant your first plant           |
 | **condition**                    | plants.count() >= 1              |
-| **track_id**                     | 1                                |
+| **track**                        | Beginners Track                  |
 | **icon**                         | NULL                             |
 | **is_seasonal**                  | false                            |
+
+## `Blossoms Gained`
+
+| **_Column name_**                | **_Example_**                    | **_Initial rule_**                                                                                | **_Description_**                  |
+| :------------------------------- | :------------------------------- | :------------------------------------------------------------------------------------------------ | :--------------------------------- |
+| **id**                           | 1                                |
+| **times_gained**                 | 1                                |
 
 ## `Relation`
 
--- a/doc/architecture/database_schemata.md
+++ b/doc/architecture/database_schemata.md
@@ -223,6 +223,7 @@ blossoms_gained }o--|| users : ""
 | **honors**                       | 0                                | 0 to infinity
 | **visits**                       | 0                                | 0 to infinity
 | **harvested**                    | 0                                | 0 to infinity, amount of plants harvested on this map
+| **version_date**                 | 2023-04-04                       | the date the snapshot for this version was taken
 | **creation_date**                | 2023-04-04                       |
 | **deletion_date**                | 2023-04-04                       |
 | **zoom_factor**                  | 100                              | value used in formula "X by X cm", e.g. 100 would mean "100 x 100 cm", range from 10 to 100000
--- a/doc/architecture/database_schemata.md
+++ b/doc/architecture/database_schemata.md
@@ -138,7 +138,7 @@ blossoms_gained {
   INT times_gained
 }
 
-maps }o--|| users : owned by
+maps }o--|| users : "owned by"
 blossoms ||--o{ blossoms_gained : ""
 blossoms_gained }o--|| users : ""
 
--- a/doc/architecture/database_schemata.md
+++ b/doc/architecture/database_schemata.md
@@ -113,6 +113,7 @@ users {
   VARCHAR experience
   VARCHAR preferences
   GEOGRAPHY location
+  INT membership_points
 }
 
 blossoms {
@@ -212,12 +213,12 @@ blossoms_gained }o--|| users : ""
 | **version_name**                 | NULL                             |
 | **is_inactive**                  | false                            |
 | **last_visit**                   | 2023-04-04                       |
-| **honors**                       | 0                                |
-| **visits**                       | 0                                |
-| **harvested**                    | 0                                |
+| **honors**                       | 0                                | | 0 to infinity
+| **visits**                       | 0                                | | 0 to infinity
+| **harvested**                    | 0                                | | 0 to infinity
 | **creation_date**                | 2023-04-04                       |
 | **deletion_date**                | 2023-04-04                       |
-| **zoom_factor**                  | 100                              |
+| **zoom_factor**                  | 100                              | | value used in formula "X by X cm", e.g. 100 would mean "100 x 100 cm", range from 10 to 100000
 | **background**                   | NULL                             |
 | **background_scale**             | NULL                             |
 | **geo_data**                     | NULL                             | | PostGis Geodata
@@ -235,6 +236,7 @@ blossoms_gained }o--|| users : ""
 | **experience**                   | {beginner, hobbyist, expert}     |
 | **preferences**                  | raised vegetable beds            |
 | **location**                     | Vienna, Austria                  |
+| **membership_points**            | 0                                | | 0 to infinity
 
 ## `Blossoms`
 
@@ -243,7 +245,7 @@ blossoms_gained }o--|| users : ""
 | **id**                           | 1                                |
 | **title**                        | Novice Gardener                  |
 | **description**                  | Plant your first plant           |
-| **condition**                    | plants.count() >= 1              |
+| **condition**                    | plants.count() >= 1              | | condition used to check if milestone is reached
 | **track**                        | Beginners Track                  |
 | **icon**                         | NULL                             |
 | **is_seasonal**                  | false                            |
@@ -253,7 +255,7 @@ blossoms_gained }o--|| users : ""
 | **_Column name_**                | **_Example_**                    | **_Initial rule_**                                                                                | **_Description_**                  |
 | :------------------------------- | :------------------------------- | :------------------------------------------------------------------------------------------------ | :--------------------------------- |
 | **id**                           | 1                                |
-| **times_gained**                 | 1                                |
+| **times_gained**                 | 1                                | | 0 to infinity
 
 ## `Relation`
 
--- a/doc/architecture/database_schemata.md
+++ b/doc/architecture/database_schemata.md
@@ -145,6 +145,27 @@ plant_detail }|--|| family : ""
 | **nutrition_demand**             | NULL                             | If "Nutritionally poor soil" in `environmental_tolerances` is given `light feeder` should be set. |
 | **preferable_permaculture_zone** | NULL                             |                                                                                                   | -1..6 (-1 should be printed as 00) |
 
+## `Maps`
+
+| **_Column name_**                | **_Example_**                    | **_Initial rule_**                                                                                | **_Description_**                  |
+| :------------------------------- | :------------------------------- | :------------------------------------------------------------------------------------------------ | :--------------------------------- |
+| **id**                           | 1                                |
+| **owner**                        | 1                                |
+| **name**                         | My Map                           |
+| **is_variant**                   | false                            |
+| **version_name**                 | NULL                             |
+| **is_inactive**                  | false                            |
+| **last_visit**                   | 2023-04-04                       |
+| **honors**                       | 0                                |
+| **visits**                       | 0                                |
+| **creation_date**                | 2023-04-04                       |
+| **deletion_date**                | 2023-04-04                       |
+| **zoom_factor**                  | 100                              |
+| **background**                   | NULL                             |
+| **background_scale**             | NULL                             |
+| **geo_data**                     | NULL                             |
+| **epgs_code**                    | NULL                             |
+
 ## `Relation`
 
 Many-to-many table to store relations between plants, genus, subfamily and family.
--- a/doc/architecture/database_schemata.md
+++ b/doc/architecture/database_schemata.md
@@ -88,9 +88,9 @@ plant_detail }|--|| family : ""
 maps {
   INT id PK
   VARCHAR name "NOT NULL"
-  BOOLEAN is_variant "NOT NULL"
+  BOOLEAN is_version "NOT NULL"
   VARCHAR version_name
-  BOOLEAN is_inactive
+  BOOLEAN is_inactive "NOT NULL"
   DATE last_visit
   INT honors
   INT visits
@@ -98,7 +98,6 @@ maps {
   DATE creation_date
   DATE deletion_date
   INT zoom_factor
-  BYTEA background
   INT background_scale
   GEOGRAPHY geo_data
   INT epsg_code
@@ -110,10 +109,11 @@ users {
   DATE free_until
   VARCHAR app_language
   DATE member_since
-  VARCHAR experience
+  INT[] member_years
+  EXPERIENCE experience
   VARCHAR preferences
   GEOGRAPHY location
-  INT membership_points
+  INT[] membership_points
 }
 
 blossoms {
@@ -133,9 +133,16 @@ enum_tracks {
   VARCHAR Expert_Track
 }
 
+enum_experience {
+  VARCHAR beginner
+  VARCHAR advanced
+  VARCHAR expert
+}
+
 blossoms_gained {
   INT id PK
   INT times_gained
+  DATE gained_date
 }
 
 maps }o--|| users : "owned by"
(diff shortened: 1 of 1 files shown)


doc/architecture/database_schemata.md
```diff
@@ -149,3 +202,57 @@ subfamily }|--|| family : ""
...
+| **_Column name_**                | **_Example_**                    | **_Initial rule_**                                                                                | **_Description_**                  |
+| :------------------------------- | :------------------------------- | :------------------------------------------------------------------------------------------------ | :--------------------------------- |
+| **id**                           | 1                                |
+| **owner_id**                     | 1                                |
+| **name**                         | My Map                           |
+| **is_variant**                   | false                            |
+| **version_name**                 | NULL                             |
+| **is_inactive**                  | false                            |
+| **last_visit**                   | 2023-04-04                       |
+| **honors**                       | 0                                |
```
badnames, line 217: What are the constraints for this value? 
Can it grow up to infinity? 
Can it be negative? 

If this information is already present in some other doc, please just add a link to prevent information from going out of sync.

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,61 @@ species }|--|| genus : ""
...
+  INT epsg_code
+}
+
+users {
+  INT id PK
+  VARCHAR nc_uid
+  DATE free_until
+  VARCHAR app_language
+  DATE member_since
+  VARCHAR experience
```
badnames, line 115: I think this should be an enum, since we only ever assign fixed values here.

badnames (approved):
Looks good thus far. 
The only thing that is missing for me are a few descriptions of database constraints (foreign keys, minima and maxima for numbers).

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,62 @@ species }|--|| genus : ""
 genus }|--|| subfamily : ""
 subfamily }|--|| family : ""
 
+maps {
+  INT id PK
+  VARCHAR name "NOT NULL"
+  BOOLEAN is_variant "NOT NULL"
+  VARCHAR version_name
+  BOOLEAN is_inactive
```
markus2330, line 95: Why not "NOT NULL"?

doc/architecture/database_schemata.md
```diff
@@ -149,3 +203,58 @@ subfamily }|--|| family : ""
...
+| **id**                           | 1                                |
+| **owner_id**                     | 1                                |
+| **name**                         | My Map                           |
+| **is_variant**                   | false                            |
+| **version_name**                 | NULL                             |
+| **is_inactive**                  | false                            |
+| **last_visit**                   | 2023-04-04                       |
+| **honors**                       | 0                                | | 0 to infinity
+| **visits**                       | 0                                | | 0 to infinity
+| **harvested**                    | 0                                | | 0 to infinity
```
markus2330, line 220: Please add description.

doc/architecture/database_schemata.md
```diff
@@ -149,3 +203,58 @@ subfamily }|--|| family : ""
 | **is_tree**                      | true                             | Set of Herbaceous/Woody (woody) AND life cycle (perennial)                                        |
 | **nutrition_demand**             | NULL                             | If "Nutritionally poor soil" in `environmental_tolerances` is given `light feeder` should be set. |
 | **preferable_permaculture_zone** | NULL                             |                                                                                                   | -1..6 (-1 should be printed as 00) |
+
+## `Maps`
+
+| **_Column name_**                | **_Example_**                    | **_Initial rule_**                                                                                | **_Description_**                  |
```
markus2330, line 209: Initial rule is probably not relevant outside plants.

doc/architecture/database_schemata.md
```diff
@@ -149,3 +203,58 @@ subfamily }|--|| family : ""
...
+| **membership_points**            | 0                                | | 0 to infinity
+
+## `Blossoms`
+
+| **_Column name_**                | **_Example_**                    | **_Initial rule_**                                                                                | **_Description_**                  |
+| :------------------------------- | :------------------------------- | :------------------------------------------------------------------------------------------------ | :--------------------------------- |
+| **id**                           | 1                                |
+| **title**                        | Novice Gardener                  |
+| **description**                  | Plant your first plant           |
+| **condition**                    | plants.count() >= 1              | | condition used to check if milestone is reached
```
markus2330, line 250: Is this JavaScript?

markus2330 (changes requested):
Great job, looking forward to an updated version!

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,62 @@ species }|--|| genus : ""
 genus }|--|| subfamily : ""
 subfamily }|--|| family : ""
 
+maps {
+  INT id PK
+  VARCHAR name "NOT NULL"
+  BOOLEAN is_variant "NOT NULL"
```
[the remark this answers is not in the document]
ThorbenStaufer: There are in fact two use cases, one for layer alternatives and an additional one for map versioning. Those two database fields are designed with map versioning in mind. Is this use case no longer planned?

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,62 @@ species }|--|| genus : ""
...
+  VARCHAR version_name
+  BOOLEAN is_inactive
+  DATE last_visit
+  INT honors
+  INT visits
+  INT harvested
```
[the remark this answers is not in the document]
ThorbenStaufer: The number of times plants were harvested on this map. Used for map statistics.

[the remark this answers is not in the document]
markus2330: Thanks, please add to description in the table.

doc/architecture/database_schemata.md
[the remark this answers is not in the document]
markus2330: Exactly, both use cases should be considered. Versions are of the whole map, variants are of the layers. See also glossary.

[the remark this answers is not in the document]
markus2330: Only some layers have variants, and only some layers can be copied, see use cases in the Non-functional Constraints. Note that there are two layer use cases not yet merged. #140 and #222

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,62 @@ species }|--|| genus : ""
...
+  DATE creation_date
+  DATE deletion_date
+  INT zoom_factor
+  BYTEA background
```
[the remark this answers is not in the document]
ThorbenStaufer: Maybe I misunderstood something, but I always read the base layer as the layer were I would upload a picture of my (hypothetical) garden to be the basis for all further planning including the optional geodata to add it to the Nextcloud Geo Map. With alternatives for this layer allowed, what would stop me from also uploading a picture of a park in New Zealand? My map would no longer represent a single location. The base layer also affects all work on all the other layers since a flower planted on top of a hill in the garden could suddendly be sitting in a pond in the alternative base layer.

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,62 @@ species }|--|| genus : ""
...
+users {
+  INT id PK
+  VARCHAR nc_uid
+  DATE free_until
+  VARCHAR app_language
+  DATE member_since
+  VARCHAR experience
+  VARCHAR preferences
+  GEOGRAPHY location
+  INT membership_points
```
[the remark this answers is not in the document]
ThorbenStaufer: Do we need to persist the amount of points a user gathered for every year or only the current score? For the latter I would have proposed to reset the db field automatically every year to 0.

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,62 @@ species }|--|| genus : ""
...
+  DATE last_visit
+  INT honors
+  INT visits
+  INT harvested
+  DATE creation_date
+  DATE deletion_date
+  INT zoom_factor
+  BYTEA background
+  INT background_scale
+  GEOGRAPHY geo_data
```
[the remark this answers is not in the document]
ThorbenStaufer: As far as I understood @NurramoX, geo data in this case is the location of the whole map.

doc/architecture/database_schemata.md
ThorbenStaufer: Not yet specified as it relies on where we want to check the conditions for the gain blossoms use case. It would make sense to check it in the backend but I have to look into Rust to understand in which format we would have to save these conditions.

doc/architecture/database_schemata.md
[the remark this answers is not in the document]
ThorbenStaufer: Then I will rename these fields to versions.

doc/architecture/database_schemata.md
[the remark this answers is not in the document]
markus2330: > Maybe I misunderstood something, but I always read the base layer as the layer were I would upload a picture of my (hypothetical) garden to be the basis for all further planning

This is correct.

> including the optional geodata

Probably we have different understandings of "geodata" means. The base layer contains:

- orientation
- real length of lines etc. for georeferencing. This may or may not be an absolute position in the real world.

> to add it to the Nextcloud Geo Map.

The Nextcloud Map feature is mainly to indicate where users are. I don't think we should use the data from the base layer there. But you are right: it would be nice if also the maps could be shown there eventually, so it is a good idea to allow the user to add a coordinate for their map (but independent of the base layer functionality).

> With alternatives for this layer allowed, what would stop me from also uploading a picture of a park in New Zealand?

Nothing, if you want to copy some elements of the park in New Zealand, this is exactly how you would do it.

> My map would no longer represent a single location.

Exactly, that is why the coordinates of the map should be in a separate field and not retrieved from the base layer.

> The base layer also affects all work on all the other layers since a flower planted on top of a hill in the garden could suddendly be sitting in a pond in the alternative base layer.

When you upload alternative images, you basically only change the scale/orientation of the image, not of all the elements you already placed. This was poorly described in the use case, I fixed it in #251.

doc/architecture/database_schemata.md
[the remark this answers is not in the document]
markus2330: It should be an array of the points the user collected in every year for administrative purposes.

doc/architecture/database_schemata.md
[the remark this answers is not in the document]
markus2330: Then this is fine, please add this to the description.

doc/architecture/database_schemata.md
markus2330: Please describe this shortly how you want to do it before you start implementing. We should keep this simple.

doc/architecture/database_schemata.md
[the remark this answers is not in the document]
ThorbenStaufer: > When you upload alternative images, you basically only change the scale/orientation of the image, not of all the elements you already placed.

That is exactly my point. If the elements placed stay static but the underlying picture changes, then there would be the possibility that elements are in a place on the alternative image were they logically would not make sense, e.g. coordinates of plant X stay the same but in the alternative image you can see a body of water in exactly that place. You would need to either place the elements in locations that make sense for both base images or create an alternative layer for each base layer. And that would be confusing for the users, especially in a shared map scenario, to keep track which (alternative) layer is supposed to be used with which base image.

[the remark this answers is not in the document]
markus2330: It is a complicated matter. I updated #251 again to clarify the different needs:

1. The base image might be of the same place, but is just a different picture/drawing/... In this case you basically only stretch and rotate the image until it fits into the same polygon as before (Alternative scenario 1.)
2. The user actually wants to change where all the elements in the database are (the image is right and not the currently placed elements). This is **not supported** as alternative. So there can be only one geometry for every element on every layer.

We should also talk with Luciano about this, especially the georeferencing. Base layer is actually one of our first use cases. But most likely not the part of changing the geometry.

doc/architecture/database_schemata.md
```diff
@@ -149,3 +203,58 @@ subfamily }|--|| family : ""
...
+| **nc_uid**                       | 1                                | | Nextcloud ID
+| **free_until**                   | 2023-04-04                       |
+| **app_language**                 | English                          |
+| **member_since**                 | 2023-04-04                       |
+| **experience**                   | {beginner, hobbyist, expert}     |
+| **preferences**                  | raised vegetable beds            |
+| **location**                     | Vienna, Austria                  |
+| **membership_points**            | 0                                | | 0 to infinity
+
+## `Blossoms`
```
aidnurs, line 243: Should we add "blossoms" to the glossary?

doc/architecture/database_schemata.md
```diff
@@ -149,3 +203,58 @@ subfamily }|--|| family : ""
...
+
+| **_Column name_**                | **_Example_**                    | **_Initial rule_**                                                                                | **_Description_**                  |
+| :------------------------------- | :------------------------------- | :------------------------------------------------------------------------------------------------ | :--------------------------------- |
+| **id**                           | 1                                |
+| **title**                        | Novice Gardener                  |
+| **description**                  | Plant your first plant           |
+| **condition**                    | plants.count() >= 1              | | condition used to check if milestone is reached
+| **track**                        | Beginners Track                  |
```
aidnurs, line 251: Could you please update the description of this field? For me as a reviewer, the field is not self describing without any prior knowledge

aidnurs (changes requested):
Well done! I think that is a good suggestion for the schemas, I understand how complicated it is to construct something before actually implementing it. Everything looks reasonable for me! The only thing that I would suggest avoiding is specifying the details/construction of the map for the time being e.g. `epsg_code` or `geo_data`. This could help us merge your initial suggestion as it is and come back later once the map research is completed

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,62 @@ species }|--|| genus : ""
...
+enum_tracks {
+  VARCHAR Beginners_Track
+  VARCHAR Seasonal_Track
+  VARCHAR Completionist_Track
+  VARCHAR Expert_Track
+}
+
+blossoms_gained {
+  INT id PK
+  INT times_gained
```
markus2330, line 140: What would be also interesting is to know which tracks are too easy/too difficult, so statistics how many users are how far ahead in which track. Is this derivable from this data?

doc/architecture/database_schemata.md
ThorbenStaufer: I will do that today. Thanks for reminding me, I meant to do that a while ago already.

doc/architecture/database_schemata.md
ThorbenStaufer: It should be possible to create a detailed report from the blossoms_gained data via Joins with the blossoms and the users table. SQL alone might already work, even though querying the data and then processing them externally would probably be simpler.

I will also include a timestamp attribute in the table, so that we can also add a time constraint to this report.

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,62 @@ species }|--|| genus : ""
...
+  INT honors
+  INT visits
+  INT harvested
+  DATE creation_date
+  DATE deletion_date
+  INT zoom_factor
+  BYTEA background
+  INT background_scale
+  GEOGRAPHY geo_data
+  INT epsg_code
```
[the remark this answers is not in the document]
NurramoX: Yeah will do!

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,69 @@ species }|--|| genus : ""
 genus }|--|| subfamily : ""
 subfamily }|--|| family : ""
 
+maps {
+  INT id PK
+  VARCHAR name "NOT NULL"
+  BOOLEAN is_version "NOT NULL"
```
markus2330, line 93: Somehow it is always a version, the question is if it is the latest. Do you maybe mean this?

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,69 @@ species }|--|| genus : ""
...
+  VARCHAR version_name
+  BOOLEAN is_inactive "NOT NULL"
+  DATE last_visit
```
markus2330, line 96: more importantly, we need a date of the version

markus2330 (changes requested):
Please rebase so that we can merge and then let @NurramoX create a follow up PR.

doc/architecture/database_schemata.md
ThorbenStaufer: I imagined an entry with is_version set to false as the root, similar to the branch head in git. When loading a version of the map, the system would query for all maps with the same root name and is_version set to true. Adopting certain elements of a map version would then be no different than overwriting attributes in/linked to the root map entry.

It may also work with having everything be technically a version, but we would need to be sure that we don't need some kind of a root/base map version.

doc/architecture/database_schemata.md
ThorbenStaufer: At first I thought we could reuse creation_date, but you are right. A dedicated version_date field makes sense.

doc/architecture/database_schemata.md
markus2330: doc/architecture/glossary.md defines version as a snapshot of a map, so everything in maps is actually versions of maps? Maybe we don't need to know what the first/last (root/current) is, as it should be determinable via the date of the version?

doc/architecture/database_schemata.md
markus2330: Yes, creation of the map vs. creation of the specific version.

markus2330: The creation date doesn't change, even if the oldest versions get deleted.

doc/architecture/database_schemata.md
markus2330: Yes, probably we should make the versions a separate table?

markus2330: Sounds like a good plan!

ThorbenStaufer: @NurramoX I ping you so you also know about this conceptual change with to what table the layers should refer.

doc/architecture/database_schemata.md
```diff
@@ -85,6 +85,70 @@ plant_detail }|--|| genus : ""
 plant_detail }|--|| subfamily : ""
 plant_detail }|--|| family : ""
 
+maps {
+  INT id PK
+  VARCHAR name "NOT NULL"
+  BOOLEAN is_version "NOT NULL"
+  VARCHAR version_name
+  BOOLEAN is_inactive "NOT NULL"
```
markus2330, line 93: Also make a date for inactivity, is more flexible (can be used to trigger deletion after some period) and consistent to deletion.

doc/architecture/database_schemata.md
```diff
@@ -145,6 +209,65 @@ plant_detail }|--|| family : ""
...
+| **geo_data**                     | NULL                             | PostGis Geodata, location of the map
+| **epsg_code**                    | NULL                             | PostGis EPSG Code
+
+## `Users`
+
+| **_Column name_**                | **_Example_**                    | **_Description_**                  |
+| :------------------------------- | :------------------------------- | :----------------------------------|
+| **id**                           | 1                                |
+| **nc_uid**                       | 1                                | Nextcloud ID
+| **free_until**                   | 2023-04-04                       | has free membership until the given date
```
markus2330, line 240: avoid the word "free", see wording in use case.

doc/architecture/database_schemata.md
```diff
@@ -85,6 +85,70 @@ plant_detail }|--|| genus : ""
...
+  INT id PK
+  VARCHAR nc_uid
+  DATE free_until
+  VARCHAR app_language
+  DATE member_since
+  INT[] member_years
+  EXPERIENCE experience
+  VARCHAR preferences
+  GEOGRAPHY location
+  INT[] membership_points
```
markus2330, line 116:
```suggestion
  INT[] permacoins
```

doc/architecture/database_schemata.md
```diff
@@ -85,6 +85,70 @@ plant_detail }|--|| genus : ""
...
+  VARCHAR title
+  VARCHAR description
+  VARCHAR condition
+  TRACKS track
+  BYTEA icon
+  BOOLEAN is_seasonal
+}
+
+enum_tracks {
+  VARCHAR Beginners_Track
```
markus2330, line 130: should be all lower case, see #155

doc/architecture/database_schemata.md
```diff
@@ -145,6 +209,65 @@ plant_detail }|--|| family : ""
 | **nutrition_demand**             | NULL                             | If "Nutritionally poor soil" in `environmental_tolerances` is given `light feeder` should be set. |
 | **preferable_permaculture_zone** | NULL                             |                                                                                                   | -1..6 (-1 should be printed as 00) |
 
+## `Maps`
+
+| **_Column name_**                | **_Example_**                    | **_Description_**                  |
+| :------------------------------- | :------------------------------- | :----------------------------------|
+| **id**                           | 1                                |
+| **owner_id**                     | 1                                |
+| **name**                         | My Map                           |
```
markus2330, line 218: which characters are allowed?

markus2330 (changes requested):
Great work, thanks for your endurance! Unfortunately there are further changes required and a few fixes (versions) are not yet done. I think it is better if we get it right from the start.

doc/architecture/database_schemata.md
```diff
@@ -87,8 +87,62 @@ species }|--|| genus : ""
 genus }|--|| subfamily : ""
 subfamily }|--|| family : ""
 
+maps {
+  INT id PK
+  VARCHAR name "NOT NULL"
+  BOOLEAN is_variant "NOT NULL"
+  VARCHAR version_name
```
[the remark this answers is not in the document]
ThorbenStaufer: I was waiting for @NurramoX to finish the layer design so that we could plan the versions table together, since it concerns us both. But I can try to staple the general map table and the individual layers together by myself if you want.

doc/architecture/database_schemata.md
ThorbenStaufer: You are right. 'gained_date' should probably be an array of dates, so we can keep track of every time it was earned.

markus2330:
Let us continue (versions etc.) in a follow-up.

PR #289, opened 2023-04-16
backend testing strategy

kitzbergerg:
## Basics

## Checklist

      (not in the PR description)

## Review

### DIFF
--- a/doc/architecture/glossary.md
+++ b/doc/architecture/glossary.md
@@ -126,8 +126,8 @@
   In the backend unit tests can be found in the `src/` directory. They are to be put in separate modules and annotated with `#[cfg(test)]` (that way they are only compiled when actually running tests).  
   Example: Test an individual function for specific behaviour.
 - integration test:  
-  Used to test whether different parts of the code work together correctly.  
-  In the backend integraton tests can be found in the `tests/` directory (which is separate from the `src/` directory).  
+  Used to test if different parts of the code work together correctly.  
+  In the backend integration tests can be found in the `tests/` directory (which is separate from the `src/` directory).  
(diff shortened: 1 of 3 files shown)


kitzbergerg:
I would make actual implementation changes in a separate PR.

Benjaminimal (approved):

Bushuo (approved):

badnames (approved):

kitzbergerg:
@markus2330 Since we don't actually use 'integration tests' according to our current definition I think we should redefine those.

I would propose to call integration tests all tests we put in the `src/test/` directory which are supposed to test multiple layers (therefore the test we currently have would be called integration tests from now on). 
Unit tests on the other hand should be put directly in the files where their related functions are and they are only supposed to test individual modules/functions (we currently wouldn't have any unit tests, but we might in the future).

markus2330:
> Since we don't actually use 'integration tests' according to our current definition (and probably never will) I think we should redefine those.

Yes.

> I would propose to call integration tests all tests we put in the src/test/ directory which are supposed to test multiple layers (therefore the test we currently have would be called integration tests from now on).
Unit tests on the other hand should be put directly in the files where their related functions are and they are only supposed to test individual modules/functions (we currently wouldn't have any unit tests, but we might in the future).

Sounds like a good idea. Probably we should call them "backend integration tests", though as there will be probably also "frontend integration tests" (testing frontend with available backend).

markus2330:
Closing was by accident :wink:

kitzbergerg:
@markus2330 Unless the latest commit introduced unwanted changes this can be merged. The actual implementation is part of #302. If there are other changes necessary I would add them over there.

PR #292, opened 2023-04-17
Load map only after click

### LINKED ISSUE #255: load external map only on click or scroll event
Instead of the external map a screenshot should be shown. Below there should be text like "On click an external resource will be loaded."

ThorbenStaufer:
The geo map on the landing page will now be dynamically loaded only if the user clicks on the placeholder image.
Resolves #255.

## Basics

## Checklist

      (not in the PR description)

## Review


frontend/src/features/landing_page/components/GeoMap.tsx
```diff
@@ -47,4 +50,29 @@ export const GeoMap = () => {
...
+    content = (
+      <div className="text-center" id="placeholder">
+        <img
+          src="/permaplant-map-placeholder.png"
+          alt="PermaplanT Map"
+          className="mb-2 hover:cursor-pointer"
+          onClick={handleClick}
+        />
+        <i>{t('geomap:hint')}</i>
+        <br />
```
Bushuo, line 72:
```suggestion
        <span className="my-2 flex-1 shrink-0 grow text-sm italic">{t('geomap:hint')}</span>
```
The `<i>` tag should not be used here. It has the semantic meaning of reading something in a different voice or mood such that it is different from the rest of the text.
See here for the exact spec https://html.spec.whatwg.org/#the-i-element

Also, why do we need the `<br>`?

I think smaller would look a little better. The other styles are such that it does not overflow the parent on smaller screens.

frontend/src/features/landing_page/components/GeoMap.tsx
```diff
@@ -47,4 +50,29 @@ export const GeoMap = () => {
       {markers}
     </MapContainer>
   );
+
+  function handleClick() {
+    setIsVisible(true);
+  }
```
Bushuo, line 56: I would prefer to inline this
```tsx
onClick={() => setIsVisible(true)}
```
or putting it to the top, above all the markup.

I think grouping the code like that makes it easier to read.

frontend/src/features/landing_page/components/GeoMap.tsx
```diff
@@ -47,4 +50,29 @@ export const GeoMap = () => {
...
+
+  if (isVisible) {
+    content = map;
+  } else {
+    content = (
+      <div className="text-center" id="placeholder">
+        <img
+          src="/permaplant-map-placeholder.png"
+          alt="PermaplanT Map"
+          className="mb-2 hover:cursor-pointer"
```
Bushuo, line 68:
```suggestion
      <div className="flex h-full w-full flex-col text-center" id="placeholder">
        <img
          src="/permaplant-map-placeholder.png"
          alt="PermaplanT Map"
          className="mb-2 h-full w-full object-cover hover:cursor-pointer"
```

Making the img span the whole container and adding `object-size: cover` enhances the image on smaller screens.
It crops the edges and places it in the center. Without shrinking it down.
The flex container is so that the span below can grow to the size it needs and the img shrinks a little.

frontend/src/features/landing_page/components/GeoMap.tsx
```diff
@@ -47,4 +50,29 @@ export const GeoMap = () => {
...
   );
+
+  function handleClick() {
+    setIsVisible(true);
+  }
+
+  let content;
+
+  if (isVisible) {
+    content = map;
```
Bushuo, line 61:
```suggestion
  let content;

  if (isVisible) {
    content = map;
```
Instead of doing this we could add an early return.
```ts

if (!isVisible) {
  // the placeholder from below. 
  return placeholder;
}

return map;

File: frontend/src/features/landing_page/components/GeoMap.tsx [hunk 5]
```diff
@@ -47,4 +50,29 @@ export const GeoMap = () => {
...
+          className="mb-2 hover:cursor-pointer"
+          onClick={handleClick}
+        />
+        <i>{t('geomap:hint')}</i>
+        <br />
+      </div>
+    );
+  }
+
+  return <>{content}</>;
```
Bushuo, line 77:
```suggestion
  return content;
```

We don't need the fragment here.

Bushuo (changes requested):
Everything looks good, just a few styling changes and readability enhancement.

markus2330:
@absurd-turtle can you have a look here? Needs a few fixes and rebase.

markus2330 (changes requested):
Map better should be in Nextcloud, so that we can update it without a release. (@absurd-turtle)

frontend/src/features/landing_page/components/GeoMap.tsx
ThorbenStaufer: I forgot to delete the `<br>`. Originally there was a dedicated button to load the map, but I later replaced it with the whole map div as the button.

frontend/src/features/landing_page/components/GeoMap.tsx
ThorbenStaufer: This is my first experience with Tailwind, so I still don't know fully when and where I can and should add what classes.
But thank you for the tips, I will try to also think of them in future work.

frontend/src/features/landing_page/components/GeoMap.tsx
ThorbenStaufer: You're right, I totally didn't think of that.

markus2330:
Would be great to get that finished today, then we can include it in the release.

ThorbenStaufer:
Already added the suggested changes to the code. Only thing left is fetching the map from Nextcloud.
I can try to read into it.

Bushuo (approved):

markus2330:
Works really nice! A big thanks, the feature made it into the release! :rocket:

PR #302, opened 2023-04-20
#39 cleanup of unit tests

kitzbergerg:
## Basics

## Checklist

      (not in the PR description)

## Review

- [ ] I've tested the code


kitzbergerg:
This would be an initial draft of the unit test cleanup. It seems a lot neater like this although there now is the problem that we cannot (we didn't do that before either) validate the state of the database after the test is done. Ideally we should test if something was actually removed from the database after e.g. calling the delete endpoint.
Maybe we can create a struct wrapping the helper functions.

@Benjaminimal @Bushuo @badnames Feel free to leave suggestions about how we want to handle that. You can also remove yourself as reviewers and unsubscribe if you don't fell like adding something (especially since this is in somewhat early stages)

kitzbergerg:
> Maybe we can create a struct wrapping the helper functions so we have a
> 
> ```rust
> let test = MyTestStruct::init_db(|conn|...);
> test.run_test(|app| ...);
> test.validate_db(|conn| ...);
> ```

I'll give up on this variant. This soon gets really annoying with Traits, Futures and the likes (as `app` in this example is not a struct but just Trait bounds and I would have to define them correctly which is quite difficult).

kitzbergerg:
@markus2330 The cleanup in that sense is now done.
I believe we should still add validation of the database's state after the tests are run. Also there are some tests missing e.g. delete for seeds. Should this be included in this PR?

markus2330 (approved):
LGTM, thank you for the cleanup!

markus2330:
Unfortunately needs rebase. You can also add the new test case(s) here.

kitzbergerg:
@markus2330 I just rebased. I would prefer to merge it as soon as possible to avoid additional merge conflicts. This PR affects every new testcase we will add in the future, so every new test case added in some other branch would lead to a conflict if this is not merged. I didn't think about that.

So I would prefer adding new tests in a separate PR to get this merged as soon as possible.

markus2330:
Thank you for merging the master branch here! (It is different to rebasing, though.)

Issue #217, opened 2023-04-04
prepare scraper for CI


markus2330:
To do  #211 it actually it would be better if plants table gets cleaned within the scraper, so that the scraper does in one transaction:

- [x] remove everything in plants table
- [x] add all the plants

Further features:

- [x] removing&inserting of all plants must be within one transaction
- [x] the merge tool should be extended to merge permapeople and reinsaat (next to practicalplants)
- [x] as extra feature, only the changed plants should be fetched from permapeople (full fetch should be available, too)
- [x] as the practicalplants output is always the same, it doesn't need to be redone on the CI
- [x] insert order into DB should stay stable

# EDIT 09.04.2023

Since the merging and insertion of the data will be the same for all the scrapers, this issue will be a logical followup of #230 and #231

markus2330:
@aidnurs as there maybe is confusion about the transaction, I summarized above all the features the scraper finally should have (should be nothing new, everything was already said somewhere):

- [ ] removing&inserting of all plants must be within one transaction
- [ ] the merge tool should be extended to merge permapeople and reinsaat (next to practicalplants)
- [ ] as extra feature, only the changed plants should be fetched from permapeople (full fetch should be available, too)
- [ ] as the practicalplants output is always the same, it doesn't need to be redone on the CI (probably nothing to be done by you, as long as parsing practicalplants stays as an extra command)

markus2330:
Last but not least (was already the case so far):

- [ ] insert order into DB should stay stable

This might actually need consideration for permapeople and reinsaat, maybe even en extra table with all known plants so far?

aidnurs:
Hey guys @markus2330 @yvonnemarkl ,

which column names would you prefer on the overlapping values e.g. `light_requirement` from permapeople vs. `sun` from practicalplants?

markus2330:
We will add this to the same PR describing the mapping.

PR #307, opened 2023-04-23
document backend architecture resolves #294 resolves #247

### LINKED ISSUE #247: graceful shutdown timeout
On the usual signals (TERM, HUP, ...) the backend should gracefully shut down:

- [ ] already started API calls should be finished
- [ ] new API calls should be terminated with "retry later"
- [ ] signals and behavior should be documented
- [ ] there should be a test case that demonstrates that API calls get finished (e.g. introduce an API call that simply waits a second or so)

### LINKED ISSUE #294: document backend architecture
Should answers questions like:

    - Both `entity` and `dto` exists as file and directory in the `model` package.
      Shouldn't they both be a regular package of their own?
    - Both `entity` and `dto` split each struct and it's function implementations across two files.
      Wouldn't it make more sense to put struct and functions together into one file for each?
    - Why is `dto` part of the service layer?
      As this is the way we represent data in the api it fits better as subpackage of `controller`.
    - Intersection of layers.
      The service layer should be the one acting between controllers and database.
      Why do controllers get the database connection pool as argument when this would better fit the service layer?
      Why does the database layer handle mapping from entity to dto when this should belong in the service layer?

kitzbergerg:
## Basics

## Checklist

      (not in the PR description)

## Review

- [ ] I've tested the code


backend/src/model/entity/plants_impl.rs
```diff
@@ -6,10 +6,7 @@ use diesel_async::{AsyncPgConnection, RunQueryDsl};
 use crate::db::function::array_to_string;
 use crate::db::pagination::Paginate;
 use crate::model::dto::{Page, PageParameters, PlantsSearchParameters};
-use crate::{
```
markus2330, line 9: Probably better to put this into a separate PR to not cause conflicts with #217.

doc/architecture/backend.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Structure
+
+The backend is split using a 3-layer architecture with controller, service and persistence layer.  
+Controller and service layer can be found in their respective folders.  
+The persistence layer is currently part of `model/entity/`.
+
+### Controller
+
+The controller layer contains all endpoints of the application.
```
markus2330, line 11: Please first clarify what endpoints are in our context.

doc/architecture/backend.md
```diff
@@ -0,0 +1,56 @@
...
+The backend is split using a 3-layer architecture with controller, service and persistence layer.  
+Controller and service layer can be found in their respective folders.  
+The persistence layer is currently part of `model/entity/`.
+
+### Controller
+
+The controller layer contains all endpoints of the application.  
+The actual routes are defined in `config/routes/` while the controller layer only contains the actual implementation of the endpoints.
+
+Actix passes the data the application is acting upon (in our case a pool of connections to the database) to the endpoints.
```
markus2330, line 14: Can you draw with (mermaid) how the data is passed.

doc/architecture/backend.md
```diff
@@ -0,0 +1,56 @@
...
+
+Actix passes the data the application is acting upon (in our case a pool of connections to the database) to the endpoints.  
+We then 'forward' the pool to the service layer where a connection is retrieved from the pool.  
+The persistence layer then uses that connection to make calls to the database.
+
+The endpoints are automatically documented using [utoipa](https://github.com/juhaku/utoipa) which can generated OpenAPI documentation from code.
+
+### Service
+
+The service layer is responsible for handling our business logic as well as mapping entities and DTOs.
```
markus2330, line 22:
```suggestion
The service layer is responsible for handling our business logic as well as mapping entities to DTOs.
```

doc/architecture/backend.md
```diff
@@ -0,0 +1,56 @@
...
+### Persistance
+
+The persistance layer is responsible for making connections to the database and retrieving/updating the state of the database.
+
+### Model
+
+The `model/` folder contains the data PermaplanT is acting upon.  
+Entities are shared with the database using the ORM [diesel](https://github.com/diesel-rs/diesel).  
+DTOs are shared with the frontend using [typeshare](https://github.com/1Password/typeshare).  
+The mapping of entities to DTOs occurs in the service layer.
```
markus2330, line 33: Already written above.

```suggestion
```

markus2330 (changes requested):
Great job, so good to finally have this! A few clarifications are needed and a imho one or two pictures (should be easy to do with mermaid) would make everything much clearer.

backend/src/model/entity/plants_impl.rs
kitzbergerg: I'll leave it in for now. If there are conflicts I'll create a separate PR. Its not that much work anyways.

doc/architecture/backend.md
```diff
@@ -0,0 +1,56 @@
...
+
+The files `entity.rs` and `dto.rs` in `model` contain the actual structs.
+That way you have a quick overview of what the data looks like without having to navigate multiple files.  
+The actual implementation (mappings, ...) is separate from the structs to reduce the size of the files.
```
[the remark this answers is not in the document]
kitzbergerg: It is actually in the `dto/` folder. Just the structs themselfs are in a separate folder from their actual implementation.

doc/architecture/backend.md
```diff
@@ -0,0 +1,56 @@
...
+## Structure
+
+The backend is split using a 3-layer architecture with controller, service and persistence layer.  
+Controller and service layer can be found in their respective folders.  
+The persistence layer is currently part of `model/entity/`.
+
+### Controller
+
+The controller layer contains all endpoints of the application.  
+The actual routes are defined in `config/routes/` while the controller layer only contains the actual implementation of the endpoints.
```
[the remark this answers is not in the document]
kitzbergerg: I used links for now. Although they probably won't work with mdbook, so I need to think about a soultion there.

doc/architecture/backend.md
kitzbergerg: I will do it in the glossary.

book.toml
```diff
@@ -7,3 +7,9 @@ title = "PermaplanT"
 
 [preprocessor.generate-summary]
 get_chapter_name_from_file = true
+
+[preprocessor.mermaid]
+command = "mdbook-mermaid"
+
+[output.html]
+additional-js = ["mermaid.min.js", "mermaid-init.js"]
```
markus2330, line 15: Please put these files in a sub-folder of doc. We want to keep top-level clean.

doc/architecture/backend.md
```diff
@@ -0,0 +1,68 @@
...
+## Code documentation
+
+The code documentation of the backend can be built using `cargo doc --open`.
+
+You can find a more detailed explanation of which modules do what there.
+
+## API documentation
+
+The API documentation can be viewed by running the backend with `cargo run` and then navigating to <http://localhost:8080/doc/api/swagger/ui/>.  
+It is automatically built using [utoipa](https://github.com/juhaku/utoipa).
```
markus2330, line 12: This should be rather at the end. Start with structure to give an overview what it is about.

doc/architecture/backend.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Structure
+
+The backend is split using a 3-layer architecture with controller, service and persistence layer.
+
+```mermaid
+graph LR;
+    Controller<-->Service;
+    Service<-->Persistence;
```
markus2330, line 21: Should be named like our modules are named (Persistence -> Db?). Do we really don't have a single point where a Controller interacts with Persistence? What about our "config" and "model"?

doc/architecture/backend.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+The controller can be found in [controller/](/backend/src/controller/) and service layer can be found in [service/](/backend/src/service/).  
+The persistence layer is part of [model/entity/](/backend/src/model/entity/).
+We should move this into [db/](/backend/src/db/) or create a new module like `persistance/`.
```
markus2330, line 26: Only write how it should be and create an issue for changes.

mermaid.min.js
markus2330, line 1: Must be listed in .reuse/dep5

markus2330 (changes requested):
Much better! Can we maybe have our main README.md as architecture/README.md as this is the default page that opens in mdbook? (maybe simply a symlink?)

Can you try to build mdbook in our pipeline that it does not get broken again? (E.g. when creating a folder) Probably it is simply adding one more `sh ""`...

doc/architecture/backend.md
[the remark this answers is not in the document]
markus2330: It would be enough if the link checker knows that these files exists. I created #321. Imho it is okay if you cannot click on them when reading the mdbook.

book.toml
kitzbergerg: I will put them directly into doc. I cannot create a subfolder in doc, otherwise the pre-processor would include that folder in the book.

doc/architecture/backend.md
kitzbergerg: I don't agree with that. The architecture itself is different from how it is implemented/how the code looks. The individual sections then describe where the code can be found, but is the layer should still be called Persistence. This is more about naming convention so that developers easily understand the architecture.

The controller layer should never actually interact with the persistence layer and I don't think they do in our code. The model folder simply represents our data, but has not much to do with the actual 3-layer architecture. We use DTOs and entities and that way they are separated. 
Config is the configuration of the server. It has nothing to do with the 3 layers (although there might be code in there that touch all layer, but again that is fine as the controller and persistence layer themselves don't interact).

markus2330:
jenkins build please

doc/architecture/backend.md
```diff
@@ -0,0 +1,70 @@
...
+Unit tests can be found in the modules they are supposed to test.
+
+## Graceful Shutdown
+
+Actix by default handles shutdowns (see [here](https://actix.rs/docs/server/#graceful-shutdown) for reference).  
+As soon as the signal `SIGTERM` is issued to the backend no new connections will be accepted.
+
+Still running connection will be finished or terminated after 5sec.  
+API requests should ideally take no longer than a few seconds, with a 5-second limit for logging and optimization purposes.
+Implementing this limit improves convenience for both developers (termination of endless loops) and continuous integration (faster restart on hanging API calls), with minimal complexity in the code.
```
markus2330, line 59: :+1: (itemize would be a bit better from formatting perspective)

markus2330 (changes requested):
Great job, I know documentation of architecture is not easy :sweat: Only one last point.

doc/architecture/backend.md
```diff
@@ -0,0 +1,67 @@
...
+Tests are split into unit and integration tests (see [here](/doc/tests/) for reference).
+
+Integration tests can be found in [test/](/backend/src/test/).  
+Unit tests can be found in the modules they are supposed to test.
+
+## Graceful Shutdown
+
+Actix by default handles shutdowns (see [here](https://actix.rs/docs/server/#graceful-shutdown) for reference).  
+As soon as the signal `SIGTERM` is issued to the backend no new connections will be accepted.
+Still running connection will be finished or terminated after 30sec.
```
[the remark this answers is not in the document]
markus2330: Done now, thx! :1st_place_medal:

doc/architecture/backend.md
kitzbergerg: Does it even make sense then to have a diagram? 
The modules themselves don't really interact with each other, but the code that runs inside them does. At least I have no idea how I would connect `config` or `db` to `service` or `controller`. Then there is the difference between the module `db` and persistence as I defined it. The module `db` simply contains a bit of utility functions for the database, but the actual calls to the database are made from the `model/entity/` folder.
So the question is how is module defined in arc42? Is it the structure of the code in the filesystem (in which case Rust doc handles it) or the structure of the running system (then I don't think I should describe the module `db` as it isn't the one interacting with the system)?

If I should describe the structure/modules wouldn't it be better to do something like the following:
```
config: Contains the servers configuration
controller: Contains the API endpoints
service: Contains the service layer
...
```
but then that is exactly the same as the code doc Rust generates.

kitzbergerg: Maybe the intent of arc would be more something like:

```mermaid
graph LR;
    Plants<-->Plantings;
    Seed;
```

That way the actual components are described. But then our system isn't complicated enough for something like that to make much sense. I don't even think any of those actually interact with each other.

markus2330: > Does it even make sense then to have a diagram?

Yes, to illustrate dependencies a diagram is perfect.

> The modules themselves don't really interact with each other,

Of course they do. The black box view shows if we use another module from any module, i.e., how the depend on each other. If an actual interaction takes place (calling functions) or only macros etc. are used does not really matter.

And we need to be honest with ourself, we need to either document or fix the deps we have.

> The module db simply 

Please fix the PR to contain such information.

> So the question is how is module defined in arc42?

See above: like in every architectural description (it is not specific to arc42): It needs to show dependencies between modules.

> structure of the running system (then I don't think I should describe the module db as it isn't the one interacting with the system)?

For the running system (the order of interactions) there is the runtime diagram: https://docs.arc42.org/section-6/

But this doesn't make much sense to be drawn for backend only, it should show interactions between frontend and backend.

I hope @Bushuo can jump in here. I crated #326 for the next step.

> but then that is exactly the same as the code doc Rust generates.

Obviously, architecture is only a high-level description of the code.

doc/architecture/backend.md
```diff
@@ -0,0 +1,75 @@
...
+### Controller
+
+The controller layer contains all endpoints of the application.  
+The actual routes are defined in [config/routes.rs](/backend/src/config/routes.rs) while the controller layer only contains the actual implementation of the endpoints.
+
+When an endpoint gets called Actix clones an internally stored pool of connections to the database and passes it to the endpoint to be used.  
+We then 'forward' the pool to the service layer where a connection is retrieved from the pool.  
+The persistence layer then uses that connection to make calls to the database.
+
+The endpoints are automatically documented using [utoipa](https://github.com/juhaku/utoipa) which can generated OpenAPI documentation from code.
```
absurd-turtle, line 34:
```suggestion
The endpoints are automatically documented using [utoipa](https://github.com/juhaku/utoipa) which can generate OpenAPI documentation from code.
```

doc/architecture/glossary.md
```diff
@@ -40,6 +40,9 @@
   - the time it takes for the page to become fully interactive.
 - The Document Object Model (DOM)
   - the data representation of the objects that comprise the structure and content of a document on the web.
+- endpoints
+  - The API endpoints that can be called in the backend.  
+    Their documentation can be view using swagger-ui (see [here](/doc/architecture/backend.md#api-documentation) for reference).
```
absurd-turtle, line 45:
```suggestion
    Their documentation can be viewed using swagger-ui (see [here](/doc/architecture/backend.md#api-documentation) for reference).
```

absurd-turtle (commented):
Looks great! :)

markus2330 (approved):
Great job!

PR #351, opened 2023-05-12
320 authentication client for frontend

absurd-turtle:
## Basics

## Checklist

      (not in the PR description)
- [x] I mentioned [every code or binary](/.reuse/dep5) not directly written or done by me in [reuse syntax](https://reuse.software/)

## Review

- [ ] I've tested the code


markus2330:
Target date to merge this PR: Tuesday 14:00.

backend/.env.sample
```diff
@@ -2,3 +2,7 @@
 DATABASE_URL=postgres://permaplant:permaplant@127.0.0.1/permaplant
 BIND_ADDRESS_HOST=127.0.0.1
 BIND_ADDRESS_PORT=8080
+
+# OAuth2 (other URLs will be fetched from this URL)
+AUTH_DISCOVERY_URI=http://localhost:8081/realms/PermaplanT/.well-known/openid-configuration
+AUTH_CLIENT_ID=PermaplanT
```
markus2330, line 8: Why would we want to configure this?

backend/README.md
```diff
@@ -13,6 +16,8 @@
 - `DATABASE_URL` is the Connection [URI](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING) to your PostgreSQL database
 - `BIND_ADDRESS_HOST` defines the host on which the server will run on
 - `BIND_ADDRESS_PORT` defines the port on which the server will run on
+- `AUTH_DISCOVERY_URI` the .well-known endpoint of the auth server (see [RFC 8414](https://www.rfc-editor.org/rfc/rfc8414.html#section-2) for more detail)
+- `AUTH_CLIENT_ID` the client id the frontend should use to log in
```
markus2330, line 20: Let us hard-code this, it also makes the description in "Setup Keycloak" again correct.

backend/README.md
```diff
@@ -45,19 +50,44 @@ LC_ALL=C diesel migration run
...
+Try to execute a request (it should return error 401).  
+Click `Authorize`.
 
-Example requests:
-
-- `curl localhost:8080/api/plants`
+> Use the `authorizationCode` auth flow.  
+> Enter client_id `swagger-ui` (client_secret is empty) and click `Authorize`.  
+> Enter user credentials (username: `test`, password: `test`).  
+> You should now be able to execute a request in Swagger.
```
markus2330, line 90: Why would this be a note? It is an essential part to test server using Swagger

```suggestion
- Use the `authorizationCode` auth flow.
- Enter client_id `swagger-ui` (client_secret is empty) and click `Authorize`.
- Enter user credentials (username: `test`, password: `test`).
- You should now be able to execute a request in Swagger.
```

backend/README.md
```diff
@@ -45,19 +50,44 @@ LC_ALL=C diesel migration run
...
 cargo run
 ```
 
-## Usage
+### Test server using Swagger
 
-Now the server is running and will start listening at <http://localhost:8080/> (or whichever port you specified in `.env`).
+Go to <http://localhost:8080/doc/api/swagger/ui/>.  
+Try to execute a request (it should return error 401).  
+Click `Authorize`.
```
markus2330, line 85:
```suggestion
- Go to <http://localhost:8080/doc/api/swagger/ui/>.
  If you try to execute a request now it should return error 401.
- Click `Authorize`.
```

devel/keycloak/docker-compose.yml
markus2330, line 1: File doesn't seem to be used nor documented. If you add a README.md what it is for, you can move it to `doc/setups` or similar.

devel/nextcloud/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+    build:
+      context: .
+      dockerfile: ../nginx/Dockerfile
+    ports:
+      - "443:443"
+      - "80:80"
+    restart: always
+
+  db:
+    image: mariadb:10.6
```
markus2330, line 18: Wouldn't it make more sense to use PostgreSQL everywhere?

frontend/src/App.tsx
```diff
@@ -1,37 +1,57 @@
...
+    switch (auth.activeNavigator) {
+      case 'signinSilent':
+        toast('Signing you in...');
+        break;
+      case 'signoutRedirect':
+        toast(`Signing you out...`);
+    }
+    console.log('authenticated: ', auth.isAuthenticated);
+    if (auth.isAuthenticated) {
+      toast(`Hello ${auth.user?.profile.sub}`);
```
markus2330, line 39: This comes quite often and prints a long hexadecimal number in my case. It is also missing translation.

frontend/src/App.tsx
```diff
@@ -1,37 +1,57 @@
...
+
+  // authentication notifications
+  useEffect(() => {
+    console.log('auth loading: ', auth.isLoading);
+    if (auth.error) {
+      toast(`Oops... ${auth.error.message}`);
+    }
+    switch (auth.activeNavigator) {
+      case 'signinSilent':
+        toast('Signing you in...');
```
markus2330, line 32: Please translate all strings.

frontend/src/features/auth/components/LoginButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+};
+
+export const LoginButton = () => {
+  const auth = useAuth();
+
+  if (auth.isAuthenticated) {
+    return (
+      <div>
+        <SimpleButton className="pl-6 pr-0" onClick={() => void auth.removeUser()}>
+          Log out
```
markus2330, line 16: Please translate all strings.

frontend/src/features/landing_page/api/getImages.ts
```diff
@@ -0,0 +1,34 @@
+import { createNextcloudAPI } from '@/config/axios';
+
+export const getPhotos = async (): Promise<Array<string>> => {
```
markus2330, line 3: Great to see that but it would probably be better to have this in a different PR. It definitely needs some documentation what it is for.

frontend/src/routes/index.ts
```diff
@@ -43,6 +44,11 @@ const routes: Routes = {
     path: '/maps',
     title: 'Map Overview',
   },
+  [Pages.Login]: {
+    component: Login,
+    path: '/login',
```
markus2330, line 49: Why make it configurable if it is hard-coded after all. VITE_REDIRECT_URI can probably be removed, too. @Bushuo already worked on an issue where such a useless config was removed.

markus2330 (changes requested):
Great job, everything worked for me. Only some simplifications and documentation is needed.

markus2330:
One quite nasty bug: when the frontend cannot reach keycloak or backend anymore, the whole landing page gets replaced by:

image

Expected behavior:
User simply gets logged out by timeout and still sees the landing page.

Can also be fixed by follow-up issue, but please create the follow-up issues first.

backend/.env.sample
kitzbergerg: There are different clients configured for different instances e.g. https://www.permaplant.net or https://pr.permaplant.net/.
Those need different CLIENT_IDs.

backend/README.md
kitzbergerg: see comment above

backend/README.md
kitzbergerg: No idea how that happened, but it wasn't intended.

frontend/.env.sample
```diff
@@ -1,4 +1,14 @@
...
+# Env variables must have the prefix VITE_ for Vite to include them.
+# More information about env variables with Vite can be found here: https://vitejs.dev/guide/env-and-mode.html
 
 VITE_BASE_API_URL=http://localhost:8080
+
+VITE_NEXTCLOUD_URI=http://localhost:8081
+
+# authentication
+# The URL of the OIDC/OAuth2 provider.
+# VITE_AUTHORITY=http://0.0.0.0:8081/realms/PermaplanT
```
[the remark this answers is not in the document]
kitzbergerg: @absurd-turtle This whole section should probably be removed from the frontend it isn't necessary any more.

frontend/.env.sample
```diff
@@ -1,4 +1,14 @@
...
+# client application's identifier as registered with the OIDC/OAuth2 provider.
+# VITE_CLIENT_ID=PermaplanT
```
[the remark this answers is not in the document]
kitzbergerg: env vars should be removed entirely from frontend. See above

frontend/.env.sample
```diff
@@ -1,4 +1,14 @@
...
+# The redirect URI of your client application to receive a response from the OIDC/OAuth2 provider.
+# VITE_REDIRECT_URI=http://localhost:5173/login
```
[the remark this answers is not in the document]
kitzbergerg: env vars should be removed entirely from frontend. See above

devel/keycloak/docker-compose.yml
absurd-turtle: I used it to run my local instance of keycloak for testing and development purposes.
I don't know if we will need it later on or if we want to test against the production instance or some dedicated online development instance. 
The same goes for Nextcloud.
If we want to have some local dev containers for these services I suggest we improve on these docker files.
As of now keycloak has to be configured after creating the instance. This could be improved by having the needed settings in a build script so not every dev will have to concern themselves with setting up keycloak.
Maybe we can also reuse what @lukashartl wrote for the production environment.

frontend/.env.sample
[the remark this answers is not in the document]
markus2330: perfect!

frontend/src/features/landing_page/api/getImages.ts
absurd-turtle: Yes, as mentioned in some other PR. The purpose was for testing requests with credentials against the NextcloudAPI.
I can add a comment mentioning this, so we can improve on it in a dedicated issue.

markus2330: Yes, write that at the top of the file. Then everything is clear.

frontend/src/routes/index.ts
absurd-turtle: Yes, correct.
In fact the whole route, component and env variable is rendered useless with the recent changes.
I will remove it.

absurd-turtle:
> One quite nasty bug: when the frontend cannot reach something (keycloak or backend?) anymore, the whole landing page gets replaced by:
> 
> image
> 
> Expected behavior: User simply gets logged out by timeout and still sees the landing page.
> 
> Can also be fixed by follow-up issue, but please create the follow-up issues first.

This happens when the backend isn't reachable.
Good point though. There is no reason why the landing page should not be displayed. I will look into it.

frontend/src/features/landing_page/api/getImages.ts
absurd-turtle: There is probably a better location for it though.

absurd-turtle: I created a separate issue and branch for this: #364 
and moved the functions there

backend/.env.sample
kitzbergerg: I would ask @lukashartl to add that there then. Maybe even write in the doc how the client_id will be called per deployment.

@markus2330 Another option would be to hardcode the client_id (although I don't really like hardcoding values like that) and instead use different realms.
The current solution would mean we can/have to use the same users for https://dev.permaplant.net, https://pr.permaplant.net and https://www.permaplant.net. With different realms users are separate and you could for example only allow us developers on https://pr.permaplant.net. 

Again I would refer to @lukashartl for that as he is the one setting up Keycloak.

devel/keycloak/docker-compose.yml
absurd-turtle: The only reason was because the documentation in the backend simply did not exist at the time and I didn't really know where to put the docker-compose file.
In general I prefer using docker-compose as the commands can get quite long when adding volumes and such.

backend/.env.sample
markus2330: Yes, I have to agree it would be better to have the same users on all three instances. But what is the purpose of the client_id then? Is it possible to have a test user for only one of the instances?

kitzbergerg: The have different origins. In Keycloak you have to define the root URL and also where you are allowed to redirect users to for each client.
E.g. https://dev.permaplant.net is a different root URL than https://pr.permaplant.net. So if you try to log in via a client that has https://dev.permaplant.net as root URL, but you make that call from https://pr.permaplant.net (you try to reuse a client) Keycloak wouldn't allow that. I think it would technically be possible to define multiple allowed redirect URLs, but that seems very hacky.

absurd-turtle:
> > One quite nasty bug: when the frontend cannot reach something (keycloak or backend?) anymore, the whole landing page gets replaced by:
> > image
> > Expected behavior: User simply gets logged out by timeout and still sees the landing page.
> > Can also be fixed by follow-up issue, but please create the follow-up issues first.
> 
> This happens when the backend isn't reachable. Good point though. There is no reason why the landing page should not be displayed. I will look into it.

Unfortunately there is no easy fix for this.
The AuthProvider expects the correct configuration. We can delay the initialization of the AuthProvider and wait for the configuration from the backend and then intitialize it.
If the backend is down the AuthProvider never receives the config and we can return an error message.
This was our first approach. This leads to the problem that nothing except an error message is displayed because the AuthProvider can't be initialized.

We tried to fix this by passing a default config.
Then wait for the config from the backend (until it is online again) and then rerendering the AuthProvider component.
This can achieved by providing a key prop:

`const config = data ?? defaultConfig;`
`return <AuthProvider key={config.authority} {...config}>{children}</AuthProvider>;`

Now the problem is that once the configuration has changed and the AuthProvider is reloaded the login does not work anymore.

markus2330:
@Bushuo @badnames @buenaflor can you take a look here?

devel/keycloak/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+version: "2"
```
markus2330, line 1: Please move to `doc/setups`, `devel` could be anything.

Bushuo (approved):

markus2330 (changes requested):
Last fix: please also add how to use public instance.

markus2330:
jenkins build please

backend/.env.sample
markus2330: The documentation also lacks which are valid client IDs for auth.permaplant.net (PermaplanT didn't work for me)

Changelog should contain both that:

- now the config gets fetched from backend
- that auth is active

backend/README.md
```diff
@@ -52,8 +52,19 @@ cargo install typeshare-cli
 
 6. Start Keycloak
 
+via docker run:
```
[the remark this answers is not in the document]
lukashartl: I have created the client as requested.

* Client ID = `localhost`
* Root URL = `http://localhost:5173/`
* Valid Redirect URLs = `/*`
* Web Origins = `+`

@absurd-turtle , @markus2330 could someone pls verify whether this works? 😄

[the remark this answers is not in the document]
markus2330: Thank you, following .env works flawlessly for me (including self-registration):

```
DATABASE_URL=postgres://permaplant:permaplant@127.0.0.1/permaplant
BIND_ADDRESS_HOST=localhost
BIND_ADDRESS_PORT=8080
AUTH_DISCOVERY_URI=http://auth.permaplant.net/realms/PermaplanT/.well-known/openid-configuration
AUTH_CLIENT_ID=localhost
```

@absurd-turtle or @kitzbergerg please do the last updates, so that we can finally merge this.

lukashartl (approved):

backend/.env.sample
kitzbergerg: The configuration was still set up for the local Keycloak instance and there exists a client ID PermplanT as described in the setup steps.
I changed the sample to point to the dev instance on auth.permaplant

kitzbergerg (approved):

backend/.env.sample
kitzbergerg: The client IDs needed for development are in the documentation (in the setup steps).
With changelog you mean add it in the next meeting notes? Or do we have an actual changelog? If so I couldn't find it.

absurd-turtle:
> > > One quite nasty bug: when the frontend cannot reach something (keycloak or backend?) anymore, the whole landing page gets replaced by:
> > > image
> > > Expected behavior: User simply gets logged out by timeout and still sees the landing page.
> > > Can also be fixed by follow-up issue, but please create the follow-up issues first.
> > 
> > 
> > This happens when the backend isn't reachable. Good point though. There is no reason why the landing page should not be displayed. I will look into it.
> 
> Unfortunately there is no easy fix for this. The AuthProvider expects the correct configuration. We can delay the initialization of the AuthProvider and wait for the configuration from the backend and then intitialize it. If the backend is down the AuthProvider never receives the config and we can return an error message. This was our first approach. This leads to the problem that nothing except an error message is displayed because the AuthProvider can't be initialized.
> 
> We tried to fix this by passing a default config. Then wait for the config from the backend (until it is online again) and then rerendering the AuthProvider component. This can achieved by providing a key prop:
> 
> `const config = data ?? defaultConfig;` `return <AuthProvider key={config.authority} {...config}>{children}</AuthProvider>;`
> 
> Now the problem is that once the configuration has changed and the AuthProvider is reloaded the login does not work anymore.
> 
> I have created a seperate issue for this here: [#365](https://github.com/ElektraInitiative/PermaplanT/issues/365)

This issue is now fixed in [#365](https://github.com/ElektraInitiative/PermaplanT/issues/365)

markus2330 (approved):
@kitzbergerg mdbook checks unfortunately still fails.

Other than that we can merge from my point of view.

kitzbergerg:
@markus2330 should be done

markus2330:
pr.permaplant.net should be tested too

markus2330:
Finally done! Great job!

backend/.env.sample
markus2330: Yes, the Changelog in each meeting note is our actual changelog. For this week #377

PR #357, opened 2023-05-15
feat: add plants

buenaflor:
After investigation the feature drag and drop turns out to be more non-trivial than expected. Since the tool bar and the map live in different components (which is a good thing), it is very difficult to synchronize a drag and drop feature as it would require a lot of introduced complexity for a feature that might have a diminishing return.

A simple click on the plant in the toolbar which adds it to the map is in my opinion much more straightforward and allows us to focus on other critical features without adding unnecessary complexity.

## Basics

## Checklist

      (not in the PR description)

## Review

- [ ] I've tested the code

### DIFF
--- a/frontend/src/config/i18n/de/plantSearch.json
+++ b/frontend/src/config/i18n/de/plantSearch.json
@@ -1,4 +1,4 @@
 {
-  "dnd": "Drag and drop Pflanzen",
+  "dnd": "Pflanzen hinzufügen",
   "placeholder": "Pflanzen suchen..."
 }
--- a/frontend/src/config/i18n/en/plantSearch.json
+++ b/frontend/src/config/i18n/en/plantSearch.json
@@ -1,4 +1,4 @@
 {
-  "dnd": "Drag and drop plants",
+  "dnd": "Add plants",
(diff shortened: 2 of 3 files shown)


markus2330:
> A simple click on the plant in the toolbar which adds it to the map

Clicking instead of drag&drop is actually a very good idea because it allows us to also draw an area of plants, e.g., a strawberry field by using click&move when placing the strawberry.

> adding unnecessary complexity

Is the drag&drop the thing that adds the complexity? Essential is that people see which plants work with which plants and on which spots. So when planting the plants (also by clicking), they need to see:

1. what they are currently planting, ideally next to the cursor like often done for drag&drop but also okay if it is highlighted where they clicked before.
2. which places on the map are suitable
3. connection (lines) to the related plants nearby on the map.

"1." and "2." are a rather static features, it could also be activated via click but "3." is the crucial feature which needs dynamic updates depending on where the mouse cursor is. And this will probably not easy to be implemented.

badnames:
@markus2330 @buenaflor Is this PR still active?
If not we should propably close it.

markus2330:
Yes, it is already implemented.

Issue #270, opened 2023-04-13
document way to get current language


markus2330:
> Additionally, I would try to use the Accept-Language header to infer the language to be used.
How do we determine the language in the frontend?
Is it a manual setting for the user or do we get it from the browser settings?

It is both. If the user loads the page for the first time the language is guessed by the browser settings, with the fallback being english. Then this setting is persisted into local storage.
I think Accept-Language alone will not work, because changing the language on the page does not change this header. I can look into it in the evening though.

_Originally posted by @Bushuo in https://github.com/ElektraInitiative/PermaplanT/issues/268#issuecomment-1506640156_

markus2330:
@Bushuo wrote:
> ```ts
> axios.interceptors.request.use((config) => {
>   config.headers['Accept-Language'] = i18next.resolvedLanguage;
>   return config;
> });
> ```
> 
> This would add the correct `Accept-Language` header to each fetch request.

Sounds okay to me, once we add this, we should document the headers we set for the backend.

I assign this now to @Benjaminimal as part of the search.

Benjaminimal:
Looks good!

PR #367, opened 2023-05-21
#258 plants full-text search

### LINKED ISSUE #258: Provide full text-search for plants
To improve searching for plants we want to implement a full-text search.

For details on how it should work see:
- Original discussion.
  https://github.com/ElektraInitiative/PermaplanT/pull/246#discussion_r1161097428
- Use case.
  https://github.com/ElektraInitiative/PermaplanT/blob/master/doc/usecases/polyculture/search_plants.md

Postgres full-text search docs.
https://www.postgresql.org/docs/current/textsearch.html

Diesel full-text search.
https://github.com/diesel-rs/diesel_full_text_search
https://docs.rs/diesel_full_text_search/latest/diesel_full_text_search/index.html

kitzbergerg:
This PR will implement the full-text search for plants. 

Resolves #258 

## Basics

## Checklist

      (not in the PR description)

## Review

- [ ] I've tested the code


kitzbergerg:
@markus2330 I would now need test data. The current implementation at least executes, however there is not data in the database whatsoever so it is difficult to say how it behaves. Ideal would be a script I can execute like a migration that simply inserts some plants into the database on startup.

markus2330:
- On pr.libelektra.org the testdata is available already.
- Locally you can import it easily, too. It is available under Nextcloud Database/mergedDatasets.csv for your convenience, simply copy it to the data folder of the scraper and run `npm run insert`.

kitzbergerg:
> * On pr.libelektra.org the testdata is available already.
> * Locally you can import it easily, too. It is available under Nextcloud Database/mergedDatasets.csv for your convenience, simply copy it to the data folder of the scraper and run `npm run insert`.

Got it working locally. However there seems to be a mismatch between the data inserted in the database and the structs we defined in the backend. A least the `WaterRequirement` enum is missing variants. 
@markus2330 Should I modify the enum or the import script?

kitzbergerg:
@markus2330 I implemented a simple variant of the search for the frontend and it seems quite fast. There is still a bit to do in the frontend, but I would leave that for somebody who actually knows what's going on there so I would ask you to assign somebody (or we do it in the meeting).

Apart from that I'm somewhat done (or at least I would leave it like it is until further discussion as I don't really know what else is to do here in the backend).

doc/decisions/search_plants.md
```diff
@@ -90,11 +93,25 @@ Cons:
 
 ## Decision
 
+Given the constrains and assumptions Partial Match with String Similarity (pg_trgm) seems to be the best option for us.
+
 ## Rationale
 
+As there is a relatively small number of rows and search terms performance isn't that much of an issue.
```
markus2330, line 100: If you have some data about search performance, please add it here.

doc/decisions/search_plants.md
```diff
@@ -90,11 +93,25 @@ Cons:
...
+
+The ability to handle spelling mistakes and loose matches is also great for us as plant names can be somewhat complicated to spell.
```
markus2330, line 102: :+1:

doc/decisions/search_plants.md
```diff
@@ -90,11 +93,25 @@ Cons:
...
+Generally it is seems to fit best for our usecase.
+Something like elasticsearch could cover the usecase better, however it adds a lot of complexity.
+
 ## Implications
 
--
+The following points are difficult if not impossible to implement using PostgreSQL.  
+Before trying them it should be evaluated if a different database like elasticsearch might be a better fit for this usecase.
+
+- Other columns can be matched with extra syntax (e.g. environmental fit or ecological value)
```
markus2330, line 112: Can you elaborate? We were thinking about something like `water_requirement = dry`, which basically simply adds one WHERE clause. Especially if the SQL query is text already, it should be simple to implement (if you keep the extra syntax simple).

doc/decisions/search_plants.md
```diff
@@ -90,11 +93,25 @@ Cons:
...
+- Search in hierarchy is difficult
+- Search accuracy (stop words, stemming, etc.)
```
markus2330, line 114: Is this maybe irrelevant because of the spell correction? Wouldn't e.g. apricots get a fix to find apricot?

backend/src/model/entity/plants_impl.rs
```diff
@@ -14,6 +15,42 @@ use crate::{
 use super::Plants;
 
 impl Plants {
+    /// Get the top 10 plants matching the search query.
```
markus2330, line 18: Please write in an issue which features are missing.

markus2330 (changes requested):
Great job, I really like the decision update. Please create an issue with open points (additional to frontend implementation).

doc/decisions/search_plants.md
kitzbergerg: I simply copied that from the use case Benjamin wrote (and as it isn't covered by the current solution and difficult do do I added it here).
The stemming itself might work with spell correction, however it is still a different concept.
Stop words aren't really a probelm as well as they should occur where we are searching.

I can remove this line if you want.

doc/decisions/search_plants.md
kitzbergerg: Apart from what I wrote here I can tell you from experience that knowbody would ever use a feature like that apart from the developers/designer themselves.
I have seen exactly that with use case with elasticsearch 2 times at different workplaces and in both cases no user actually used that feature (even though it was natively supported).
The reason being is that the user would have to know all column names in the database by heart or at least look at the documentation an figure out which variants to enter where.
A solution that I have commonly seen for that is to provide additional fields in the frontend where the user can enter e.g. water_requirement if they want to. However then there is no need to parse the query in the backend as it could simply be part of the query struct (although that might be difficult as well in the current situation as again we have to be careful about SQL injections).

kitzbergerg:
> I don't agree, there are several essential parts missing, e.g. pagination etc.

I just tried around with pagination and that seems to be way more difficult than anticipated, because of the raw SQL statement I'm using. What I've tried so far:
1. Simply use `.paginate()` like in seeds. Doesn't work because there is some trait bound that is wrong with diesel. Somehow diesel cannot infer what the return type should be because I'm using a raw SQL query.
2. Tried switching to non raw SQL query. This works almost however I still need the raw sql query for the `similarity()` function as diesel doesn't support it. This again results in problem although I don't exactly know why, but for whatever reason diesel doesn't seem be able to parse the `similarity() function` (or there is some other error in the way I built up that query). The problem here is more that there is basically no documentation about how to build up a query as a mix of raw SQL and diesel statement. Apart from that there might be other errors as well I haven't yet discovered.
3. Trying to use `LIMIT $2 OFFSET $3` directly in my raw SQL. That way I can use the same query params as in other endpoints. However when parsing the result of that there are again errors with diesel and some trait bound (which probably again has to do with the raw SQL query). So I can send the query using that way (I have tried it actually), but I'm not able to convert the result to the `Page` struct. It should be possible to return a version of `Page` with `total_pages` missing, however then the frontend won't know when it has to stop getting the next page. So this version is probably the one I'll go with and at some point the frontend will simply receive an empty list instead of results and that is when it will stop querying.

The simplest solution would be to ignore pagination for this kind of use case.

@markus2330 For pagination there are now the following options:
- If its ok that the frontend simply gets an empty page at some point then I will do that. 
- If there should be 'normal' pagination then this issue will take a lot longer as I probably have to either deep dive into diesel or somehow implement a diesel variant of the `similarity()` function myself (further variants of point 2)
- We rethink the way search works in the backend (although I don't know if other variants will be easier). Maybe the diesel full-text search might help, although again that will require quite some time to figure out as there is limited documentation (I had a look at https://github.com/diesel-rs/diesel_full_text_search)


Note: When I'm talking about trait bound in diesel the error looks somewhat like the following (and this is a simple and easy to understand version of the errors that diesel returns):
```
the trait bound `(Untyped, BigInt): CompatibleType<(_, i64), Pg>` is not satisfied
the following other types implement trait `CompatibleType<U, DB>`:
  (ST0, ST1)
  (ST0, ST1, ST2)
  (ST0, ST1, ST2, ST3)
  (ST0, ST1, ST2, ST3, ST4)
  (ST0, ST1, ST2, ST3, ST4, ST5)
  (ST0, ST1, ST2, ST3, ST4, ST5, ST6)
  (ST0, ST1, ST2, ST3, ST4, ST5, ST6, ST7)
  (ST0, ST1, ST2, ST3, ST4, ST5, ST6, ST7, ST8)
and 120 others
required for `PaginatedQuery<UncheckedBind<UncheckedBind<UncheckedBind<SqlQuery, &String, Text>, i32, Integer>, i32, ...>>` to implement `diesel_async::methods::LoadQuery<'_, AsyncPgConnection, (_, i64)>`
the full type name has been written to '/home/gabriel/Dev/PermaplanT/backend/target/debug/deps/backend-569138cd6411f7af.long-type-823946854615235012.txt'
```

backend/src/model/entity/plants_impl.rs
```diff
@@ -14,6 +15,42 @@ use crate::{
...
+    ///
+    /// Uses `pg_trgm` to find matches in `unique_name`, `common_name_de`, `common_name_en` and `edible_uses_en`.
+    /// Ranks them using the `PostgreSQL` function `similarity()`.
+    ///
+    /// # Errors
+    /// * Unknown, diesel doesn't say why it might error.
```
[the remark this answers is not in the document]
kitzbergerg: I removed the raw SQL so I'll leave it like it is (that way it's the same as the others).

backend/src/model/entity/plants_impl.rs
kitzbergerg: See #379

kitzbergerg:
@markus2330 Apart from `Language settings, e.g., when typing "fi" on English setting "fir" should rank higher than German "Fichte" (and the other way round)` and some other minor things regarding ranking this should be done.

For that I don't really know how to implement it. Where do I get information about the language the user is using? How much should the different language be considered when compared to the users language?
Can you maybe test if adding additional logic for ranking is even necessary? When a user is typing something in german it is quite unlikely that an english plant will appear right? Implementing additional logic regarding rankings requires larger changes again (and every other time something in the rankings will be changed) as diesel is quite inflexible regarding custom queries like that.

Apart from that. Something on master seems to have destroyed the search field in the map (or maybe it was moved somewhere)?
A kind of follow up question to that: Should @badnames implement his changes in this PR or in a separate PR?

doc/decisions/search_plants.md
markus2330: Yes, I agree. Let us remove this "special syntax" feature completely (you can remove it from the use case). At some later time (not this term) we implement a proper extended search.

markus2330:
> Simply use .paginate() like in seeds. Doesn't work because there is some trait bound that is wrong with diesel. Somehow diesel cannot infer what the return type should be because I'm using a raw SQL query.

This doesn't seem to be the case anymore (raw SQL), so you were able to fix the problem?

> Where do I get information about the language the user is using?

#270, the `Accept-Language` header should contain it.

We definitely should document this (mdbook), I created #384. Please write down which exact values it has (for English and German).

backend/src/model/entity/plants_impl.rs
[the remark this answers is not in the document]
markus2330: I don't think that this Errors description ever was good. Can't we at least say that it must be one of Diesel's errors?

kitzbergerg:
> > Simply use .paginate() like in seeds. Doesn't work because there is some trait bound that is wrong with diesel. Somehow diesel cannot infer what the return type should be because I'm using a raw SQL query.
> 
> This doesn't seem to be the case anymore (raw SQL), so you were able to fix the problem?

I fixed it by implementing a custom extension for diesel (the complicated traits usw. in the functions.rs). 

However like I mentioned diesel is quite inflexible with stuff like that. Adding a simple function like fuzzy is somewhat easy now that I understand how to do it, but combining different custom function in a way that works is really difficult. For example the current ranking using `greatest`, `similarity` and `alias` basically only works in that exact configuration (adding an additional column to the ranking algorithm or returning a different type than f32 doesn't work). So if you want a more complicated ranking algorithm it is probably a better idea to use a mix of diesel functions and raw sql (see [here](https://github.com/ElektraInitiative/PermaplanT/pull/367/commits/75dbf3958c2c61e62239e66f78e9a4bb8560ba16#diff-8f559b8f09e35e42c9ed2a0a2565339fbe41d772d6ec3f97a1c06e40e38f5f7eR37-R45) for an example). But that requires the developer to know what is going and how diesel works in a bit more detail (otherwise the query might simply fail).

So in the end it @markus2330 you have to decide if we want to keep the ranking algorithm simple (like it is now) or if language and weights (and maybe more) should be added. 
In the second case I would switch back to the raw SQL for the ranking algorithm like in the example linked, but then the code is more complex and developers need to understand SQL, diesel and their interaction in quite a lot more depth.

markus2330 (approved):
We like it very much, the search works great.

markus2330:
Please resolve the conflict.

markus2330:
jenkins build please

PR #375, opened 2023-05-23
decision: mail client

kitzbergerg:
The decision about sending mails from the backend as written in #2.

## Basics

## Checklist

      (not in the PR description)

## Review

- [ ] I've tested the code

### DIFF
--- a/doc/decisions/backend_mail_client.md
+++ b/doc/decisions/backend_mail_client.md
@@ -13,14 +13,14 @@ The typical communication when sending a mail is as follows:
 
 ### Example use case
 
-A user enters a question into the contract form on the frontend.  
+A user enters a question into the contact form on the frontend.  
 The frontend should now send a request to the backend.  
-The backend then send a request to somebody who can answer this question.
+The backend then send a mail to somebody who can answer this question.
 
 ## Constraints
 
-- Currently there doesn't seem to exist a crate to send emails without requiring a mail-server.  
-  All crates I have found require a mail-server (and then we should also use username and password).
+- Currently there doesn't seem to exist a crate to send emails without requiring a mail-server.
+  We should therefore also use username and password as credentials in the backend.
 
 ## Solutions
 
--- a/doc/decisions/backend_mail_client.md
+++ b/doc/decisions/backend_mail_client.md
@@ -1,0 +1,53 @@
+# Mail client for the backend
+
+## Problem
+
+We want to be able to send mails from the backend.
+
+The typical communication when sending a mail is as follows:
+
+1. Your email client communicates with your SMTP server.
+2. Your SMTP server communicates with the recipient's domain's DNS server to find the address of their SMTP server.
+3. Your SMTP server communicates with the recipient's SMTP server.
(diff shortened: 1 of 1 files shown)


doc/decisions/backend_mail_client.md
```diff
@@ -0,0 +1,52 @@
...
+
+[mail_smtp](https://crates.io/crates/mail-smtp) can also be used to send emails (and also requires a mail-server).
+
+## Decision
+
+We should use the crate `lettre` for sending mails.
+
+## Rationale
+
+`lettre` is as of writing this the most widely use crate I have found with 1,343,736 total downloads.
```
absurd-turtle, line 44:
```suggestion
`lettre` is as of writing this the most widely used crate I have found with 1,343,736 total downloads.  
```

absurd-turtle (approved):

markus2330 (approved):
Great! Can you give an estimation how long it will take to provide an endpoint that can send emails with that lib? (Very easy to very difficult)

markus2330:
Btw. the alternative would be that we post text in a Nextcloud chat or create a Nextcloud Deck card when a user wants to contact us. Probably this is overall easier to be done, as it won't involve an smtp setup. Of course this has the disadvantage that a "reply" button would be missing.

@absurd-turtle what do you think?

absurd-turtle:
> Btw. the alternative would be that we post text in a Nextcloud chat or create a Nextcloud Deck card when a user wants to contact us. Probably this is overall easier to be done, as it won't involve an smtp setup. Of course this has the disadvantage that a "reply" button would be missing.
> 
> @absurd-turtle what do you think?

Yes, it could be done. We would have to send a request to the backend which then makes the request to Nextcloud. The reason is that unauthenticated requests to Nextcloud are difficult to do as far as I read.
I think if we want to receive notifications per mail go directly for mail, if we want to have them in Nextcloud, send them to Nextcloud. I don't think one of them is easier to implement.

kitzbergerg:
> Great! Can you give an estimation how long it will take to provide an endpoint that can send emails with that lib? (Very easy to very difficult)

I think it is about in the middle (a bit more on the easy side). The implementation in Rust shouldn't be that difficult, it is more about coordination and configuration that is difficult (as we need 3 new config env variables, need to decide how to locally test something like this as the mail-server is required and we need access to the mail-server from the prod backend). So like I said it isn't necessarily difficult, but takes some time to think about and plan.

> Btw. the alternative would be that we post text in a Nextcloud chat or create a Nextcloud Deck card when a user wants to contact us. Probably this is overall easier to be done, as it won't involve an smtp setup. Of course this has the disadvantage that a "reply" button would be missing.

What do you mean with reply button? The way I thought about it was: A user has a question and puts it into the form on the frontend (where he also has to enter his email). The backend sends a mail with the question and the users email to somebody (e.g. @markus2330). You send a mail to the email the user entered answering the users question.
Maybe you can do some switcheroo somewhere that the mail appears to be send by the user directly instead of our technical smtp backend user (but still same concept).

PR #403, opened 2023-05-31
architecture, part II

### LINKED ISSUE #158: Rust mdBook

markus2330:
fixes #158

## Basics

## Checklist

      (not in the PR description)
      (exceptions are documented)

## Review

- [ ] I've tested the code

### DIFF
--- a/backend/src/controller/plantings.rs
+++ b/backend/src/controller/plantings.rs
@@ -7,7 +7,7 @@
 
 use actix_web::web::Query;
 use actix_web::{
-    delete, get, patch, post,
+    delete, get, post, put,
     web::{Data, Json, Path},
     HttpResponse, Result,
 };
@@ -104,7 +104,7 @@ pub async fn create(new_seed_json: Json<NewPlantingDto>, pool: Data<Pool>) -> Re
         ("oauth2" = [])
     )
 )]
-#[patch("/{id}")]
+#[put("/{id}")]
 pub async fn update(
     id: Path<i32>,
     update_seed_json: Json<UpdatePlantingDto>,
--- a/backend/src/controller/plantings.rs
+++ b/backend/src/controller/plantings.rs
@@ -7,7 +7,7 @@
 
 use actix_web::web::Query;
 use actix_web::{
-    delete, get, post, put,
+    delete, get, patch, post,
     web::{Data, Json, Path},
     HttpResponse, Result,
 };
@@ -104,7 +104,7 @@ pub async fn create(new_seed_json: Json<NewPlantingDto>, pool: Data<Pool>) -> Re
         ("oauth2" = [])
     )
 )]
-#[put("/{id}")]
+#[patch("/{id}")]
 pub async fn update(
     id: Path<i32>,
     update_seed_json: Json<UpdatePlantingDto>,
--- a/doc/guidelines/api.md
+++ b/doc/guidelines/api.md
@@ -1,0 +1,32 @@
+# REST
+
+We exclusively use:
+
+- JSON or images as content.
+- idempotent resources.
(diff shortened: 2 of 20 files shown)


backend/src/controller/plantings.rs
```diff
@@ -104,7 +104,7 @@ pub async fn create(new_seed_json: Json<NewPlantingDto>, pool: Data<Pool>) -> Re
         ("oauth2" = [])
     )
 )]
-#[patch("/{id}")]
+#[put("/{id}")]
```
Bushuo, line 107: Why not PATCH?
The actions should be as fine-grained as possible, e.g. the move action should only change x and y of a planting. 
Under the hood, it calls the update endpoint with patch, sending just
```json
{ "x": 123, "y": 123 }
```
These are the semantics of PATCH as I understand it.

PUT means that the whole dto would have to be sent. Making it impossible to undo/redo fine-grained changes.

doc/guidelines/logging.md
```diff
@@ -2,42 +2,51 @@
...
+
+To log you can simply call the macros in the `log` crate.
+
+For example to log using the debug level:
+
+```rust,ignore
+log::debug!("This is my variable: {}", some_variable);
+```
+
+As you can see the behavior is the same as it is for example for `println!()`.
```
Bushuo, line 48:
```suggestion
As you can see, it can be used in the same way as for example `println!()`.
```

Bushuo (changes requested):
I agree with everything except that we always use PUT.

backend/src/controller/plantings.rs
markus2330: If we use the idempotent subset of PATCH it is fine, too. The way you describe it, it is perfectly okay. And I agree, when the DTOs become bigger (like Plant) we definitely want to only send minimal updates.

Do we already know how the actions in the state are described? Will they be like the PATCH example you describe here?

Bushuo: > Do we already know how the actions in the state are described? Will they be like the PATCH example you describe here?
>
Yes, they look quite similar.
Though the types can not be shared.
An action in the frontend also encapsulates the behavior: 
- how to apply itself to the state.
- how to construct its inverse action.
- how to communicate itself to the backend.

Bushuo (approved):

doc/guidelines/api.md
```diff
@@ -0,0 +1,75 @@
...
+## Endpoints
+
+The endpoint paths use:
+
+- hierarchical structure
+- nouns instead of verbs
+- only plural (exception: config)
+- all endpoints need authorization (exception: config)
+- all paths below `/api`
+- layer-specific paths below `/layers/<name of layer>/` e.g. `/layers/plants/plantings`
```
Bushuo, line 24:
```suggestion
- layer-specific paths below `maps/<map_id>/layers/<name of layer>/` e.g. `/layers/plants/plantings`
```
We will need the map_id as part of every request to the layer.
Should this be encoded in the path or as part of the request body?

backend/src/controller/plantings.rs
markus2330: Yes, I see that there is more information in the state but what about having the shared DTO as sub-object?

PR #406, opened 2023-06-02
Fix/seed search

badnames:
## Basics

## Checklist

      (not in the PR description)
- [x] I mentioned [every code or binary](/.reuse/dep5) not directly written or done by me in [reuse syntax](https://reuse.software/)
- [x] I created left-over issues for things that are still to be done
      (exceptions are documented)

## Review

- [ ] I've tested the code

### DIFF
--- a/backend/src/model/entity/plants_impl.rs
+++ b/backend/src/model/entity/plants_impl.rs
@@ -40,7 +40,7 @@ impl Plants {
         // settings (0.3).
         let set_similarity_threshold = diesel::sql_query("SET pg_trgm.similarity_threshold=0.1");
         debug!("{}", debug_query::<Pg, _>(&set_similarity_threshold));
-        let _ = set_similarity_threshold.execute(conn).await;
+        let _ = set_similarity_threshold.execute(conn).await?;
 
         let query = plants::table
             .select((
--- a/backend/src/model/entity/plants_impl.rs
+++ b/backend/src/model/entity/plants_impl.rs
@@ -1,6 +1,9 @@
 //! Contains the implementation of [`Plants`].
 
-use diesel::{debug_query, dsl::sql, pg::Pg, sql_types::Float, BoolExpressionMethods, ExpressionMethods, QueryDsl, QueryResult};
+use diesel::{
+    debug_query, dsl::sql, pg::Pg, sql_types::Float, BoolExpressionMethods, ExpressionMethods,
+    QueryDsl, QueryResult,
+};
 use diesel_async::{AsyncPgConnection, RunQueryDsl};
 use log::debug;
 
@@ -35,7 +38,8 @@ impl Plants {
     ) -> QueryResult<Page<PlantsSummaryDto>> {
         // Set higher search sensitivity so that the user gets more search results than with default
         // settings (0.3).
-        let set_similarity_threshold = diesel::sql_query("SET LOCAL pg_trgm.similarity_threshold=0.1");
+        let set_similarity_threshold =
+            diesel::sql_query("SET LOCAL pg_trgm.similarity_threshold=0.1");
         debug!("{}", debug_query::<Pg, _>(&set_similarity_threshold));
         let _ = set_similarity_threshold.execute(conn).await;
 
--- a/backend/src/model/entity/plants_impl.rs
+++ b/backend/src/model/entity/plants_impl.rs
@@ -1,9 +1,6 @@
 //! Contains the implementation of [`Plants`].
 
-use diesel::{
-    debug_query, dsl::sql, pg::Pg, sql_types::Float, BoolExpressionMethods, ExpressionMethods,
-    QueryDsl, QueryResult,
-};
+use diesel::{debug_query, dsl::sql, pg::Pg, sql_types::Float, BoolExpressionMethods, ExpressionMethods, QueryDsl, QueryResult};
 use diesel_async::{AsyncPgConnection, RunQueryDsl};
 use log::debug;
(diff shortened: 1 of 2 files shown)


backend/src/model/entity/plants_impl.rs
```diff
@@ -36,6 +33,10 @@ impl Plants {
         page_parameters: PageParameters,
         conn: &mut AsyncPgConnection,
     ) -> QueryResult<Page<PlantsSummaryDto>> {
+        let set_similarity_threshold = diesel::sql_query("SET pg_trgm.similarity_threshold=0.1");
```
badnames, line 36: We might want to set this as an environment variable in the backend.
According to pg_trgm's documentation this can only be set for a specific query or for an entire session, but not for the entire database.

markus2330: No, it is fine to hard-code this.

backend/src/service/plants.rs
```diff
@@ -22,9 +22,15 @@ pub async fn find(
...
+    }
+
     let mut conn = pool.get().await?;
-    let result = match search_parameters.name {
-        Some(search_query) => Plants::search(&search_query, page_parameters, &mut conn).await?,
+    let result = match search_query {
+        Some(query) => Plants::search(&query, page_parameters, &mut conn).await?,
         None => Plants::find_any(page_parameters, &mut conn).await?,
     };
     Ok(result)
```
kitzbergerg, line 36: You can do this like the following instead:
```rust
    let mut conn = pool.get().await?;
    let result = match search_parameters.name {
        Some(query) if !query.is_empty() => {
            Plants::search(&query, page_parameters, &mut conn).await?
        }
        _ => Plants::find_any(page_parameters, &mut conn).await?,
    };
    Ok(result)
```

That uses a guard (see the if inside the match statement) and a default otherwise.

backend/src/model/entity/plants_impl.rs
```diff
@@ -36,6 +36,13 @@ impl Plants {
         page_parameters: PageParameters,
         conn: &mut AsyncPgConnection,
     ) -> QueryResult<Page<PlantsSummaryDto>> {
+        // Set higher search sensitivity so that the user gets more search results than with default
+        // settings (0.3).
+        let set_similarity_threshold =
+            diesel::sql_query("SET LOCAL pg_trgm.similarity_threshold=0.1");
+        debug!("{}", debug_query::<Pg, _>(&set_similarity_threshold));
+        let _ = set_similarity_threshold.execute(conn).await;
+
```
kitzbergerg, line 45: Is it possible to somehow avoid having to make two calls to PostgreSQL? I just had a quick look and it seems difficult do avoid this, but maybe you can think of a different solution.

kitzbergerg (changes requested):

backend/src/model/entity/plants_impl.rs
badnames: The only alternative I see is to set the value of this variable globally using an ALTER DATABASE command (e.g. as part of a migration script).

kitzbergerg: Then it should be fine like it is currently. I don't think we should invest to much time here.

backend/src/model/entity/plants_impl.rs
```diff
@@ -36,6 +36,12 @@ impl Plants {
         page_parameters: PageParameters,
         conn: &mut AsyncPgConnection,
     ) -> QueryResult<Page<PlantsSummaryDto>> {
+        // Set higher search sensitivity so that the user gets more search results than with default
+        // settings (0.3).
+        let set_similarity_threshold = diesel::sql_query("SET pg_trgm.similarity_threshold=0.1");
+        debug!("{}", debug_query::<Pg, _>(&set_similarity_threshold));
+        let _ = set_similarity_threshold.execute(conn).await;
```
kitzbergerg, line 43: One more thing in noticed. You should check for an error here. Apart from that I think it can be merged.
```suggestion
        let _ = set_similarity_threshold.execute(conn).await?;
```

kitzbergerg (approved):

markus2330:
@badnames does https://pr.permaplant.net/seeds/new work for you? For me the plant search only toggles between "Options" and "Loading"?

Solution with the setting changed permanently would be preferable. Are you sure it won't work with migrations? Afaik the migrations are executed with quite high permissions (they can alter/remove tables after all).

kitzbergerg:
> @badnames does https://pr.permaplant.net/seeds/new work for you? For me the plant search only toggles between "Options" and "Loading"?
> 
> Solution with the setting changed permanently would be preferable. Are you sure it won't work with migrations? Afaik the migrations are executed with quite high permissions (they can alter/remove tables after all).

You need to be logged in for that. So this PR is ok, but there is another issue that the page https://pr.permaplant.net/seeds/new shouldn't be reachable without being logged in. But that is a frontend problem.

backend/src/model/entity/plants_impl.rs
markus2330: Let us try to have this in migrations. But it is also fine like it is (the same connection is used, so it should be super-fast).

markus2330 (approved):
Thanks for the fixes! There is a

markus2330:
> I tried to execute the following SQL migration on my local database

Always good to let this also check by the CI. It does not happen often but sometimes somethings works on the CI and the problem is only locally.

But let us keep it like this, it is not really a problem to send this little option, even if it is for every query.

PR #404, opened 2023-05-31
Feature/sse undo redo

Bushuo:
## Basics

- [ ] Details of what you changed are in commit messages.
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildserver is happy.
- [ ] The PR is rebased with current master.

## Checklist

- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I fixed all affected decisions
- [ ] I added unit tests for my code
- [ ] I added code comments, logging, and assertions as appropriate
- [ ] I translated all strings visible to the user
- [ ] I mentioned [every code or binary](/.reuse/dep5) not directly written or done by me in [reuse syntax](https://reuse.software/)
- [ ] I created left-over issues for things that are still to be done
- [ ] Code is conforming to [our Architecture](/doc/architecture)
- [ ] Code is conforming to [our Guidelines](/doc/guidelines)
      (exceptions are documented)
- [ ] Code is consistent to [our Design Decisions](/doc/decisions)

## Review

- [ ] I've tested the code
- [ ] I've read through the whole code
- [ ] Examples are well chosen and understandable

PR #425, opened 2023-06-06
Architecture Part III

markus2330:
## Basics

- [ ] Details of what you changed are in commit messages.
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [ ] The buildserver is happy.
- [ ] The PR is rebased with current master.

## Checklist

- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I fixed all affected decisions
- [ ] I added unit tests for my code
- [ ] I added code comments, logging, and assertions as appropriate
- [ ] I translated all strings visible to the user
- [ ] I mentioned [every code or binary](/.reuse/dep5) not directly written or done by me in [reuse syntax](https://reuse.software/)
- [ ] I created left-over issues for things that are still to be done
- [ ] Code is conforming to [our Architecture](/doc/architecture)
- [ ] Code is conforming to [our Guidelines](/doc/guidelines)
      (exceptions are documented)
- [ ] Code is consistent to [our Design Decisions](/doc/decisions)

## Review

- [ ] I've tested the code
- [ ] I've read through the whole code
- [ ] Examples are well chosen and understandable

PR #427, opened 2023-06-06
Plant Layer API + DB

### LINKED ISSUE #420: implement plant DB + API
- [ ] write migrations for plant layer (correctly for maps)
- [ ] add Rust structs, see also #408
- [ ] add Rust code to implement endpoints (CRUD)

kitzbergerg:
Implementation of the Plant Layer layer including database and endpoints.
Resolves #420.

Needs the following before more detailed implementation can begin:
- #404 (for plantings) 
- #426 (for layers) 

## Basics

## Checklist

      (not in the PR description)
      (exceptions are documented)

## Review

- [ ] I've tested the code


kitzbergerg:
@markus2330 @ThorbenStaufer @Bushuo This would be an initial draft of the plants API + DB. For now a detailed review is not needed, I still need to write tests and update the doc a bit. But I would as you to review the API as a whole, maybe we want to chose different paths or something like that.
Also have a look at the plantings update API with `UpdatePlantingDto`. We could use enums like that somewhere else as well. @Bushuo I though about the actions.rs while doing that. Maybe have a look at https://serde.rs/enum-representations.html. That might give some ideas.

backend/src/model/dto/actions.rs
```diff
@@ -35,10 +35,10 @@ pub enum Action {
 #[serde(rename_all = "camelCase")]
 pub struct CreatePlantActionPayload {
     user_id: Uuid,
-    id: Uuid,
+    id: i32,
```
Bushuo, line 38: Why did you revert all the planting identifiers to `i32`?
The frontend needed those to be a `Uuid`.

For explanation, the interaction is like this:
Create planting with random uuid & add to the store to be immediately visible.
Broadcast this creation to all other clients.
The other clients create the same planting in their respective stores.

If we have integers the optimistic UI is difficult to implement, because another user might create the planting with the same id.

Bushuo (changes requested):

backend/src/model/dto/actions.rs
kitzbergerg: Then I'll change it back. Just thought it was weird that all id's are integers but that one is a uuid.

backend/src/model/dto/plantings.rs
```diff
@@ -0,0 +1,114 @@
+//! All DTOs associated with [`PlantingDto`].
```
ThorbenStaufer, line 1: Why exactly do plantings entities and DTOs need to be placed in their own module?

backend/src/config/routes.rs
```diff
@@ -31,11 +31,18 @@ pub fn config(cfg: &mut web::ServiceConfig) {
...
-                        .service(plantings::create)
-                        .service(plantings::update)
-                        .service(plantings::delete),
+                    web::scope("/{map_id}/layers")
+                        .service(layers::find)
+                        .service(layers::find_by_id)
+                        .service(layers::create)
+                        .service(layers::delete)
+                        .service(
+                            web::scope("plants/plantings")
```
ThorbenStaufer, line 40:
```suggestion
                            web::scope("plants")
```

Do we need the differentiation between plants layer and plantings on the plants layer here? Unless we want to have other, non-plantings related operations on the plants layer we can assume from context that we deal with plantings.

ThorbenStaufer (approved):
All in all no major critics from me. Just two points where I need some clarification.

markus2330 (changes requested):
Looks great, much better than before.

- [ ] we use uuid, see 
- [ ] POST for planting doesn't need scaling etc. (you cannot put it already scaled) but it will later need an area
- [ ] I didn't see the API for relations/heatmap (see doc/architecture/06runtime.md) or how is it called? In 00d9d683c5ae2b3aa84cadee462059ffa5cca52f I suggested /api/maps/{map_id}/layers/plants/placements What do you think? (Please review in #425)

backend/src/config/routes.rs
kitzbergerg: We might at some point have special operations on the plants layer itself. So `/layers` are operations on all types of layer, `/layers/plants` would be operations on the plants layer (e.g. heatmap might fit better there than under plantings), and `/layers/plants/plantings` are objects on the plants layer.

backend/src/model/dto/plantings.rs
kitzbergerg: The DTOs file is already really large so it is much easier to read and change the code when its split like this. In my opinion we should actually split up the whole DTOs and entities file and move the implementations to the structs themselves. Ideally move the db implementation to a separate module. They should be part of model.

kitzbergerg:
> * [ ]  we use uuid, see [cc4ec24](https://github.com/ElektraInitiative/PermaplanT/commit/cc4ec242bfe17cbd5e565db6fada6a306dcd5060) [Architecture Part III #425](https://github.com/ElektraInitiative/PermaplanT/pull/425)

Done

> * [ ]  POST for planting doesn't need scaling etc. (you cannot put it already scaled) but it will later need an area

I will leave it like it is for now. The frontend can set the values. If you want to have it removed please tell me exactly which values should be part of the create dto and what the default values of the other should be.

> * [ ]  I didn't see the API for relations/heatmap (see doc/architecture/06runtime.md) or how is it called? In [00d9d68](https://github.com/ElektraInitiative/PermaplanT/commit/00d9d683c5ae2b3aa84cadee462059ffa5cca52f) I suggested /api/maps/{map_id}/layers/plants/placements What do you think? (Please review in [Architecture Part III #425](https://github.com/ElektraInitiative/PermaplanT/pull/425))

I would make that part of a separate PR (tracking issue #457).

backend/src/config/routes.rs
markus2330: > We might at some point have special operations on the plants layer itself. So /layers are operations on all types of layer, /layers/plants would be operations on the plants layer

Yes.

> e.g. heatmap might fit better there than under plantings

Maybe but the heatmap is specific to individual plants, like the relations. Where do you suggest to put the heatmap?

I think we should put relations and heatmap together in an API call (as also suggested by runtime), mostly for performance reasons: to keep the number of API calls as low as possible. I wrote this now in #459 0c2c88cc62275d162c39865e579484e4bf591bf8

> and /layers/plants/plantings are objects on the plants layer.

Exactly.

Bushuo:
@kitzbergerg @ThorbenStaufer I just fixed some parts of the frontend because of the new UpdateDto. Hope you don't mind that I did it myself real quick. Everything else looks really good. Also tested it and it works.

For other testers: you need to create a map before testing the plant layer.

Bushuo (approved):

markus2330 (approved):
LGTM but CI is unhappy. @kitzbergerg can you please take a look?

kitzbergerg:
jenkins build please

kitzbergerg:
> LGTM but CI is unhappy. @kitzbergerg can you please take a look?

That is very weird. There are no errors locally and the errors in the pipeline are from a previous commit. They should actually exists. Maybe latest pipeline finished first and the result got overwritten by an older pipeline?

markus2330:
@Bushuo is there any order you would like merges to happen?

markus2330:
@kitzbergerg thx for confirming, for me bec5dbf003e12bf5a542062990f95c1f6b5623bd also builds without any problems.

markus2330:
@kitzbergerg Still broken but now the problem seems to be clear?

markus2330:
Fixed, good job!

kitzbergerg:
> @kitzbergerg Still broken but now the problem seems to be clear?

Yes. The state on master seems to result in an invalid merge (but not a merge conflict). Should be fixed now.

PR #482, opened 2023-06-19
Open map editor through map overview

ThorbenStaufer:
The map cards in the map overview page now link the user to the map editor with the respective map id in the URL.
Resolves #390.

## Basics

- [X] Details of what you changed are in commit messages.
- [X] References to issues, e.g. `close #X`, are in the commit messages.
- [x] The buildserver is happy.
- [X] The PR is rebased with current master.

## Checklist

- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I fixed all affected decisions
- [ ] I added unit tests for my code
- [ ] I added code comments, logging, and assertions as appropriate
- [ ] I translated all strings visible to the user
- [ ] I mentioned [every code or binary](/.reuse/dep5) not directly written or done by me in [reuse syntax](https://reuse.software/)
- [ ] I created left-over issues for things that are still to be done
- [ ] Code is conforming to [our Architecture](/doc/architecture)
- [ ] Code is conforming to [our Guidelines](/doc/guidelines)
      (exceptions are documented)
- [ ] Code is consistent to [our Design Decisions](/doc/decisions)

## Review

- [ ] I've tested the code
- [ ] I've read through the whole code
- [ ] Examples are well chosen and understandable

PR #488, opened 2023-06-20
Map editor connected to map

### LINKED ISSUE #481: Initialize map editor with maps layer IDs
The DB model for maps requires that data of individual layers references an layer ID instead of the map ID directly.
In order for the various API calls to include the ID of the specific layer the operation is made on, the layer IDs of the map should be loaded and stored in the frontend during map editor initialization.

ThorbenStaufer:
Follow up to #482.
Map editor will now fetch and push data to the corresponding map in the data model.
This enables the possibility to have multiple, independent maps.

Resolves #481.

## Basics

- [X] Details of what you changed are in commit messages.
- [X] The buildserver is happy.

## Checklist

- [X] I fully described what my PR does in the documentation
      (not in the PR description)
- [X] I fixed all affected documentation
- [X] I added code comments, logging, and assertions as appropriate
      (exceptions are documented)

## Review

- [ ] I've tested the code


frontend/src/features/map_planning/components/toolbar/LayerList.tsx
```diff
@@ -10,14 +10,14 @@ import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
 interface LayerListProps {
-  /** name of the layer - this is displayed on top of the slider */
-  name: LayerName;
+  /** layer that this list element is responsible for */
```
absurd-turtle, line 13:
```suggestion
  /** layer which is controlled by this list element */
```

frontend/src/features/map_planning/store/MapStoreTypes.ts
```diff
@@ -97,47 +99,58 @@ export type History = Array<Action<unknown, unknown>>;
...
-  'Terrain',
-  'Trees',
-  'Warnings',
-  'Winds',
-  'Zones',
-] as const;
+const LAYER_TYPES = Object.values(LayerType);
+
+export const TRACKED_DEFAULT_STATE: TrackedMapState = {
+  layers: {
```
absurd-turtle, line 112: I think we should refactor this to

```
  layers: LAYER_TYPES.reduce(
      (acc, layerName) => {
        acc[layerName]: {
          index: layerName,
          objects: [],
        },
        return acc
      },
      {} as TrackedLayers,
    ),
```
The spread operator can quickly slow down the code when not used carefully. The reason is that the spread operator has to iterate over the list every time it gets called.
When we mutate the variable instead we save a lot of iterations. We also don't have to be worried about side effects because the variable acc is scoped to the reduce function.

I don't think it is too much of an issue here because the size of the elements is rather small but if this is executed often it still matters and I think to try to write efficient code is still good practice.

frontend/src/features/map_planning/store/MapStoreTypes.ts
```diff
@@ -97,47 +99,58 @@ export type History = Array<Action<unknown, unknown>>;
...
+  mapId: -1,
+  selectedLayer: {
+    id: -1,
+    is_alternative: false,
+    name: 'none',
+    type_: LayerType.Base,
+    map_id: -1,
+  },
+  layers: {
+    ...LAYER_TYPES.reduce(
```
absurd-turtle, line 136: same refactor as above can be applied here

frontend/src/features/map_planning/store/MapStoreTypes.ts
ThorbenStaufer: @Bushuo can you look into this? Doesn't seem to work without some modifications and I am not really familiar with the notation.

backend/src/controller/layers.rs
```diff
@@ -6,10 +6,7 @@ use actix_web::{
...
 };
 
-use crate::{
-    config::data::AppDataInner,
-    model::dto::{LayerSearchParameters, PageParameters},
-};
+use crate::{config::data::AppDataInner, model::dto::LayerSearchParameters};
 use crate::{model::dto::NewLayerDto, service::layer};
 
 /// Endpoint for searching layers.
```
markus2330, line 12: Documentation not updated (line below doesn't make sense if we use Vec instead of Page).

frontend/src/features/map_planning/components/Map.tsx
```diff
@@ -68,36 +24,37 @@ function useMapUpdates() {
...
-      Landscape: { right: <div></div>, left: <div></div> },
-      Paths: { right: <div></div>, left: <div></div> },
-      Shade: { right: <div></div>, left: <div></div> },
-      Soil: { right: <div></div>, left: <div></div> },
-      Terrain: { right: <div></div>, left: <div></div> },
-      Trees: { right: <div></div>, left: <div></div> },
-      Warnings: { right: <div></div>, left: <div></div> },
-      Winds: { right: <div></div>, left: <div></div> },
-      Zones: { right: <div></div>, left: <div></div> },
+      [LayerType.Base]: { right: <div></div>, left: <div></div> },
```
markus2330, line 35: Rebase: there already should be a toolbar.

frontend/src/features/map_planning/store/TrackedMapStore.ts
```diff
@@ -61,13 +45,24 @@ export const createTrackedMapSlice: StateCreator<
...
-            Plant: {
-              ...state.trackedState.layers.Plant,
+            plants: {
+              ...state.trackedState.layers.plants,
               objects: plants,
             },
           },
         },
       })),
+    __resetStore() {
```
markus2330, line 55: Please add a few comments, what is this doing?

frontend/src/features/map_planning/hooks/useGetLayers.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { useQuery } from '@tanstack/react-query';
+
+/**
+ * Gets all layers for the given map id.
+ * Re-fetches data only after a certain amount of time to reduce network stress.
+ */
+export function useGetLayers(mapId: number) {
+  const data = useQuery(['layers', mapId] as const, {
+    queryFn: ({ queryKey: [, mapId] }) => getLayers(mapId),
+    staleTime: 1000 * 60 * 10, // 10 minutes
```
markus2330, line 11: use a const variable instead, see coding guidelines.

frontend/src/features/map_planning/hooks/useGetLayers.tsx
```diff
@@ -0,0 +1,15 @@
+import { getLayers } from '../api/getLayers';
+import { useQuery } from '@tanstack/react-query';
+
+/**
+ * Gets all layers for the given map id.
+ * Re-fetches data only after a certain amount of time to reduce network stress.
```
markus2330, line 6: Why refetch at all? Can't we send a SSE if a new layer is created? (please create issue if this is a bigger thing to do).

frontend/src/features/map_planning/routes/MapWrapper.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { useMapId } from '../hooks/useMapId';
+import useMapStore from '../store/MapStore';
+import { handleRemoteAction } from '../store/RemoteActions';
+import { LayerType, ConnectToMapQueryParams } from '@/bindings/definitions';
+import { baseApiUrl } from '@/config';
+import { useSafeAuth } from '@/hooks/useSafeAuth';
+import { useQuery } from '@tanstack/react-query';
+import { useRef, useEffect } from 'react';
+
+function useInitializeMap() {
```
markus2330, line 14: Please add comment(s) here.

markus2330 (changes requested):
Great work, love to see this merged soon! Please add a bit more documentation.

backend/src/controller/layers.rs
```diff
@@ -22,10 +19,9 @@ use crate::{model::dto::NewLayerDto, service::layer};
     params(
         ("map_id" = i32, Path, description = "The id of the map the layer is on"),
         LayerSearchParameters,
-        PageParameters
     ),
     responses(
-        (status = 200, description = "Search layers", body = PageLayerDto)
+        (status = 200, description = "Search layers", body = VecLayerDto)
```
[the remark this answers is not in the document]
Bushuo: The layer DTO is quite small, so there is no significant performance increase in using a page there. It would make the layer handling code in the frontend a lot more difficult.

frontend/src/features/map_planning/components/Map.tsx
Bushuo: I don't understand your comment.
We needed to change this part because we changed the string `LayerName` to the enum `LayerType`.

frontend/src/features/map_planning/hooks/useGetLayers.tsx
Bushuo: I agree.

frontend/src/features/map_planning/hooks/useGetLayers.tsx
Bushuo: Yes, this would work.
In this case, the action sent from the server could just be a notification.
Nevertheless, I would refetch the layers on receiving the notification.
Otherwise, merging the existing layers in `react-query` with the incoming layers would be complicated.

backend/src/controller/layers.rs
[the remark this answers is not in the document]
ThorbenStaufer: I agree, there is no need to paginate the layers. Also, alternatives are hidden inside a collapsable and if we are concerned about the amount of user created layer on a map, we can lazy load just the alternative layers of the layer type the user wants to uncollapse.

frontend/src/features/map_planning/components/Map.tsx
markus2330: When I open `frontend/src/features/map_planning/routes/Map.tsx` in master there is a BaseLayerRightToolbar, which seems to be missing here. So I thought that there might be a problem due to renaming this file and later changes on master.

frontend/src/features/map_planning/hooks/useGetLayers.tsx
markus2330: Yes, fetching all layers again when there was a change with the layers sounds reasonable.

frontend/src/features/map_planning/components/Map.tsx
markus2330: Probably you just need to rebase (or merge recent master) to fix this problem.

frontend/src/features/map_planning/hooks/useGetLayers.tsx
Bushuo: I added this issue: https://github.com/ElektraInitiative/PermaplanT/issues/498

Issue #342, opened 2023-05-09
Map Timeline Single Date Selection


markus2330:
This issue implements the usecase [Map Timeline Single Date Selection](https://github.com/ElektraInitiative/PermaplanT/blob/master/doc/usecases/map_timeline_single_date_selection.md)

Bushuo:
There is a library that looks promising but has other peer dependencies that I do not want to bring into the project.
https://github.com/namespace-ee/react-calendar-timeline

markus2330:
I already looked at it and didn't even find an option so that looks like how we want it. Did I miss something?

Bushuo:
I think like this, it looks quite similar.
https://codesandbox.io/s/timeline-demo-sticky-header-forked-3t7hm0?file=/CustomTimeline.jsx
![image](https://github.com/ElektraInitiative/PermaplanT/assets/9990578/50241b2b-fed6-4513-9d74-50b643358bf7)

markus2330:
It is also quite different, we don't need that the day is split in smaller units and we also don't need text per day but a graph of how many occurrences were at which day. So something like what video-players have, e.g. youtube where you see the number of watches (but we need number of changes in the map on that day).

Maybe we can reuse something from video players, and change the scale from seconds to days?

markus2330:
There are some small updates about this UC (basically what we already said) in #454

markus2330:
A probably more simple (than what we originally imagined but not more simple than a date picker) implementation of timeline could be done via server-side SVG.  See https://alexplescan.com/posts/2023/07/08/easy-svg-sparklines/ for inspiration.

We would basically write the dates with an SVG path below/around (indicating how many actions on the elements happened each day). The question is how clicking on the dates then actually would lead to the right action. Is this possible with react-svg?

If you think this makes somewhat sense, please write this as implementation hint linked the left-over of the use case.

Bushuo:
I would then prefer to use a graphing library that comes with click handling on datapoints.
One that I already used is [react-charts](https://react-charts.tanstack.com/).

markus2330:
Yeah, if you already used it the choice is obvious. Then let us forget the idea with SVG. Anyway, we won't do it for this user study.

Issue #476, opened 2023-06-19
plant attributes


markus2330:
- [ ] planting date
- [ ] harvesting date
- [ ] removal date
- [ ] rotation
- [ ] x/y position

markus2330:
I moved down the fields that are only relevant for areas.

badnames:
> seeding date (automatically set when planting)

How does this differ from planting date?

> live cycle information

Should we move this to the "maybe later section"?
The backend for this does not exist yet.

markus2330:
> How does this differ from planting date?

It differs in the case of "Vorzucht". Is the backend API already done for that?

>  Should we move this to the "maybe later section"?
> The backend for this does not exist yet.

The backend issue, however, already exists, so let us ask @horenso when it can be done.

PR #546, opened 2023-07-02
Feature/342 map timeline single date

Bushuo:
Ref: #342 

## Basics

- [X] I added a line to [/doc/CHANGELOG.md](/doc/CHANGELOG.md)
- [X] The PR is rebased with current master.
- [X] Details of what you changed are in commit messages.

## Checklist

      (not in the PR description)
- [X] I created left-over issues for things that are still to be done
      (exceptions are documented)

## Review

- [ ] I've tested the code

### DIFF
--- a/backend/migrations/2023-06-24-125250_planting_timeline_data/up.sql
+++ b/backend/migrations/2023-06-24-125250_planting_timeline_data/up.sql
@@ -1,0 +1,2 @@
+ALTER TABLE plantings ADD COLUMN add_date DATE;
+ALTER TABLE plantings ADD COLUMN remove_date DATE;
--- a/backend/migrations/2023-06-24-125250_planting_timeline_data/up.sql
+++ b/backend/migrations/2023-06-24-125250_planting_timeline_data/up.sql
@@ -1,2 +1,4 @@
 ALTER TABLE plantings ADD COLUMN add_date DATE;
 ALTER TABLE plantings ADD COLUMN remove_date DATE;
+ALTER TABLE plantings ADD COLUMN create_date DATE NOT NULL DEFAULT CURRENT_DATE;
+ALTER TABLE plantings ADD COLUMN delete_date DATE;
--- a/backend/src/model/dto.rs
+++ b/backend/src/model/dto.rs
@@ -185,7 +185,7 @@ pub struct Page<T> {
     pub total_pages: i32,
 }
 
-/// A page of loaded results in a timeline.
+/// A page of results bounded by time.
 #[typeshare]
 #[derive(Debug, Serialize, Deserialize, ToSchema)]
 #[aliases(
--- a/backend/src/model/dto.rs
+++ b/backend/src/model/dto.rs
@@ -7,6 +7,8 @@ use typeshare::typeshare;
 use utoipa::{IntoParams, ToSchema};
 use uuid::Uuid;
 
+use self::plantings::PlantingDto;
+
 use super::r#enum::{
     layer_type::LayerType, privacy_options::PrivacyOptions, quality::Quality, quantity::Quantity,
     relation_type::RelationType,
@@ -181,6 +183,21 @@ pub struct Page<T> {
     pub per_page: i32,
     /// Number of pages in total.
     pub total_pages: i32,
+}
+
+/// A page of loaded results in a timeline.
+#[typeshare]
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+#[aliases(
+    TimelinePagePlantingsDto = TimelinePage<PlantingDto>,
+)]
+pub struct TimelinePage<T> {
+    /// Resulting records.
+    pub results: Vec<T>,
+    /// The time frame start date.
+    pub from: NaiveDate,
+    /// The time frame end date.
+    pub to: NaiveDate,
 }
 
 /// The whole information of a map.
--- a/backend/src/model/dto/plantings.rs
+++ b/backend/src/model/dto/plantings.rs
@@ -126,6 +126,6 @@ pub struct PlantingSearchParameters {
     pub plant_id: Option<i32>,
     /// The id of the plants layer the planting is placed on.
     pub layer_id: Option<i32>,
-    /// A date where the plantings should exists relative to.
+    /// Plantings that exist around this date are returned.
     pub relative_to_date: NaiveDate,
 }
--- a/backend/src/model/dto/plantings.rs
+++ b/backend/src/model/dto/plantings.rs
@@ -127,5 +127,5 @@ pub struct PlantingSearchParameters {
     /// The id of the plants layer the planting is placed on.
     pub layer_id: Option<i32>,
     /// A date where the plantings should exists relative to.
-    pub relative_to_date: Option<NaiveDate>,
+    pub relative_to_date: NaiveDate,
 }
--- a/backend/src/model/dto/plantings.rs
+++ b/backend/src/model/dto/plantings.rs
@@ -1,5 +1,6 @@
 //! All DTOs associated with [`PlantingDto`].
 
+use chrono::NaiveDate;
 use serde::{Deserialize, Serialize};
 use typeshare::typeshare;
 use utoipa::{IntoParams, ToSchema};
@@ -34,6 +35,12 @@ pub struct PlantingDto {
     /// The y scale of the plant on the map.
     #[serde(rename = "scaleY")]
     pub scale_y: f32,
+    /// The date the planting was added to the map.
+    /// If None, the planting always existed.
+    pub add_date: Option<NaiveDate>,
+    /// The date the planting was removed from the map.
+    /// If None, the planting is still on the map.
+    pub remove_date: Option<NaiveDate>,
 }
 
 /// Used to create a new planting.
@@ -64,6 +71,9 @@ pub struct NewPlantingDto {
     /// The y scale of the plant on the map.
     #[serde(rename = "scaleY")]
     pub scale_y: f32,
+    /// The date the planting was added to the map.
+    /// If None, the planting always existed.
+    pub add_date: Option<NaiveDate>,
 }
 
 /// Used to differentiate between different update operations on plantings.
@@ -116,4 +126,6 @@ pub struct PlantingSearchParameters {
     pub plant_id: Option<i32>,
     /// The id of the plants layer the planting is placed on.
     pub layer_id: Option<i32>,
+    /// A date where the plantings should exists relative to.
+    pub relative_to_date: Option<NaiveDate>,
 }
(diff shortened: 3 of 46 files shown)


Bushuo:
Added Leftover issue #547

backend/src/model/entity/plantings_impl.rs
```diff
@@ -30,7 +30,22 @@ impl Planting {
...
+        query = query.filter(
+            plantings::add_date
+                .is_null()
+                .or(plantings::add_date.lt(to))
+                .and(
+                    plantings::remove_date
+                        .is_null()
+                        .or(plantings::remove_date.gt(from)),
+                ),
+        );
```
kitzbergerg, line 45: This is pretty difficult to read. Maybe you could introduce variables here to make it clearer? (Untested and bad variable names, but you get the point). It not wrong like it currently is so if my suggestion doesn't work don't spend more than a few min on it.
```suggestion
        let lower_date = plantings::add_date
                .is_null()
                .or(plantings::add_date.lt(to));
        let other_condition = plantings::remove_date
                        .is_null()
                        .or(plantings::remove_date.gt(from));
        query = query.filter(lower_date.and(other_condition));
```

backend/src/service/plantings.rs
```diff
@@ -1,14 +1,20 @@
...
 use crate::config::data::AppDataInner;
 use crate::error::ServiceError;
 use crate::model::dto::plantings::{
     NewPlantingDto, PlantingDto, PlantingSearchParameters, UpdatePlantingDto,
 };
-use crate::model::entity::plantings::Planting;
+use crate::model::dto::TimelinePage;
+use crate::model::entity::plantings::{FindPlantingsParameters, Planting};
+
+/// Time offset for loading plantings in the timeline.
```
kitzbergerg, line 16: I think its supposed to be days?
```suggestion
/// Time offset in days for loading plantings in the timeline.
```

backend/migrations/2023-06-24-125250_planting_timeline_data/up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE plantings ADD COLUMN add_date DATE;
+ALTER TABLE plantings ADD COLUMN remove_date DATE;
```
kitzbergerg, line 2: Maybe `creation_date` and `deletion_date` is better?

backend/src/model/dto.rs
```diff
@@ -182,6 +184,21 @@ pub struct Page<T> {
     pub total_pages: i32,
 }
 
+/// A page of loaded results in a timeline.
```
kitzbergerg, line 187: Not sure but maybe that sound better/is clearer?
```suggestion
/// A page of results bounded by time.
```

backend/src/model/dto/plantings.rs
```diff
@@ -116,4 +126,6 @@ pub struct PlantingSearchParameters {
     pub plant_id: Option<i32>,
     /// The id of the plants layer the planting is placed on.
     pub layer_id: Option<i32>,
+    /// A date where the plantings should exists relative to.
```
kitzbergerg, line 129: This is somewhat confusing? Do we fetch plants before the date, after the date or around the date?
Maybe it would be better to use 2 parameters instead so the frontend can control what to fetch (start_date and end_date)?

backend/src/model/entity/plantings.rs
```diff
@@ -1,10 +1,23 @@
...
+    /// The id of the plant to find plantings for.
+    pub plant_id: Option<i32>,
+    /// The id of the layer to find plantings for.
+    pub layer_id: Option<i32>,
+    /// First date in the time frame plantings are searched for.
+    pub from: NaiveDate,
+    /// Last date in the time frame plantings are searched for.
+    pub to: NaiveDate,
+}
+
```
kitzbergerg, line 20: I personally feel like you should just pass them in directly. This is not really and entity so it doesn't fit here. If you want the struct you should probably move it to where it is actually used.

kitzbergerg (changes requested):
Looks good. A few things here and there (although I was very picky).

ThorbenStaufer (approved):
Looks good to me as a first version.
Maybe add a title attribute to the "change map date" button, but isn't a high priority, since I imagine this will be replaced in the future.

backend/migrations/2023-06-24-125250_planting_timeline_data/up.sql
markus2330: We have terminology about this, see 12glossary.md

It probably should be updated to full pairs, I suggest:

- add/remove (as this PR already uses) and
- create/delete

doc/CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ Syntax: `- short text describing the change _(Your Name)_`
 - Use the correct layer ID for create and fetch actions (for plantings) in the map editor, enabling the possibility for multiple independent maps _(Paul and Thorben)_
 - Location picker map in map creation form only loads after actively pressing a button _(Thorben)_
 - Maps can be duplicated from the map overview (in PR #515) _(Thorben)_
+- A timeline date can be set and changed by a date picker. Plantings exist relative to a date. _(Paul)_
```
markus2330, line 36: Please rebase/merge master and put it to the most top section.

markus2330 (approved):
LGTM, I didn't test it though.

@kitzbergerg thx for the review!

backend/src/model/entity/plantings.rs
Bushuo: @kitzbergerg
I added them here so that the entity layer does not depend on the service layer but the other way round.
Would moving it to the impl be better?
I do not want to pass it in, because conceptually the service should handle the business logic, not the entity.

backend/src/model/entity/plantings_impl.rs
Bushuo: This worked great, thanks :)

backend/migrations/2023-06-24-125250_planting_timeline_data/up.sql
Bushuo: I added those two fields, but I did not use them. 
The delete_date seems to be relevant only once we implement hard/soft delete.

markus2330: Yes. I tried to clarify the terms in https://github.com/ElektraInitiative/PermaplanT/pull/545/files#diff-94bf40d44642ef0b5cce441bf0ae1f48175578fee52f720a298e39119ebfc9e2

backend/src/model/entity/plantings.rs
Bushuo: Another reason why I introduced a separate struct is the result error type of the entity functions.
This one produces a ServiceError.
```rs
let from = search_parameters
        .relative_to_date
        .checked_sub_days(Days::new(TIME_LINE_LOADING_OFFSET_DAYS))
        .ok_or_else(|| {
            ServiceError::new(
                StatusCode::BAD_REQUEST,
                "Could not add days to relative_to_date".into(),
            )
        })?;
```
The entities return a DieselError.

Bushuo:
@kitzbergerg 
I made some changes to the plantings tests and test module.
Shall we do this for all tests?
I think it improves the readability.
Please add your review again.

frontend/src/features/map_planning/components/timeline/Timeline.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  const onFormSubmit = ({ date }: TimelineFormData) => {
+    onSelectDate(date);
+  };
+
+  return (
+    <form className="flex items-end justify-center gap-2" onSubmit={handleSubmit(onFormSubmit)}>
+      <SimpleFormInput type="date" id="date" labelText="" register={register} />
+      <div className="w-52">
+        <SimpleButton title={t('timeline:change_date_hint')} type="submit">
```
markus2330, line 39: Please avoid the button, directly change the date when the date selector was modified.

backend/src/model/dto/plantings.rs
```diff
@@ -34,6 +35,12 @@ pub struct PlantingDto {
     /// The y scale of the plant on the map.
     #[serde(rename = "scaleY")]
     pub scale_y: f32,
+    /// The date the planting was added to the map.
+    /// If None, the planting always existed.
```
markus2330, line 39: How can you produce such a date? I think we can remove this possibility (make non-null).

If it is only legacy: don't worry, we need to reset the database anyway.

frontend/src/config/i18n/de/timeline.json
```diff
@@ -0,0 +1,4 @@
+{
+  "change_date": "Kartendatum ändern",
+  "change_date_hint": "Änder das Datum der Karte um zu sehen wie sie sich über die Zeit verändert hat und um Elemente zu diesem Zeitpunkt hinzuzufügen/ zu entfernen."
```
markus2330, line 3:
```suggestion
  "change_date_hint": "Ändere das Datum um um Elemente an diesem Tag zu sehen/hinzuzufügen/zu entfernen."
```

frontend/src/config/i18n/en/timeline.json
```diff
@@ -0,0 +1,4 @@
+{
+  "change_date": "Change map date",
+  "change_date_hint": "Change the date of the map to see how it has changed over time and add/remove elements at the given date."
```
markus2330, line 3:
```suggestion
  "change_date_hint": "Change the date to see/add/remove elements at the given day."
```

markus2330 (approved):
There is a bigger usability problem: while planting the planting information overlaps the time selector. But I think we can merge as is nevertheless and create a follow-up.

markus2330:
Rebase needed.

> There is a bigger usability problem: while planting the planting information overlaps the time selector. But I think we can merge as is nevertheless and create a follow-up.

To clarify: it is actually okay that the "dialog" is above the date selector (it is not important that the date can be changed while planting). But the user should still see the selected date. A simple solution is that we add the current selected date in the planting "dialog".

kitzbergerg (approved):
backend looks good.

backend/src/model/dto/plantings.rs
Bushuo: My idea was to enable plants that live for the whole selectable timeline. 
Like a tree that was there as long as anyone can remember.
The input in the left toolbar would allow to clear the date such that it can be null.
Default would be the current date.

markus2330: Yes, good idea. It will definitely be needed for other elements in the map (like a hill or river). :wink:

Bushuo:
@markus2330 I think this is done now. I will continue on #476 
I added the react-query dev tools which helped me a lot.
They get removed from the production bundle.

markus2330:
Thank you so much, fantastic job!

Issue #349, opened 2023-05-10
Set borders of map (frontend)


badnames:
Fence of an area within a base layer to tell PermaplanT where the borders of the background image are.

markus2330:
Thanks for creating the issue :heart:

badnames:
# Summary of changes
Because this is a bigger PR/issue than usual, I decided to summarize all new features in this comment.

![grafik](https://github.com/ElektraInitiative/PermaplanT/assets/27735433/467f535b-0bc8-4fdf-b77a-0bf681d49083)

The current map polygon is shown every time the user is one the base layer screen.
To manipulate it, there are three new buttons available in the right toolbar.

![grafik](https://github.com/ElektraInitiative/PermaplanT/assets/27735433/4998d54f-a037-414c-90e4-c32d2687aef1)
After pressing the first button, the user can move any point of the map geometry by dragging it.
This edit edit mode (as well as all other edit modes described below) can be cancelled by either pressing the 'x'-button in the status panel, switching to another layer or activating the auto scale workflow.

![grafik](https://github.com/ElektraInitiative/PermaplanT/assets/27735433/fe88f58c-3832-46ef-b655-94536f0629d3)
After pressing the second button, new points can be added to the polygon by clicking anywhere on the map.
New points will always be inserted between the nearest pair of points (measured using the sum of the distance between the new point to the first point and the distance between the new point and the second point).

![grafik](https://github.com/ElektraInitiative/PermaplanT/assets/27735433/3ce3d5bb-b1ac-470e-9bc9-c393ae5f34b1)
Using third and final button, undesired points can be deleted via a simple mouse click.

PR #566, opened 2023-07-03
Backend heatmap

### LINKED ISSUE #457: Backend plants layer placing: relations and heatmap
Implement the plants layer heatmap. Follow up to:
- #420

### LINKED ISSUE #494: Set borders of map (backend)

### LINKED ISSUE #635: document coordinate system
### Tasks

- [ ] where is 0,0 in frontend, ...
- [ ] where do x,y increase (up, left)...
- [ ] 1px = 1cm
- [ ] granularity of heatmap

### Use case

_No response_

### Related Pull request

#566

kitzbergerg:
Implementation of the backends heatmap.
Resolves #457, resolves #635, resolves #494.

## Basics

## Checklist

      (not in the PR description)
      (exceptions are documented)

## Review

- [ ] I've tested the code


backend/src/controller/map.rs
```diff
@@ -9,9 +9,31 @@ use actix_web::{
 
 use crate::config::auth::user_info::UserInfo;
 use crate::config::data::AppDataInner;
-use crate::model::dto::{MapSearchParameters, PageParameters};
+use crate::model::dto::{HeatMapQueryParams, MapSearchParameters, PageParameters};
 use crate::{model::dto::NewMapDto, service};
 
+#[utoipa::path(
+    context_path = "/api/maps",
```
markus2330, line 16: Focus on getting the API right first, so that this PR gets useful for frontend devs. Heatmaps are specific to a map and layer (other layers, like tree layer or alternative plant layers would have different heatmaps), so I would put it to `/maps/<id>/layers/plants/heatmap`.

The selected time influences the suggestions too. I assume we send this by the parameters? @Bushuo

markus2330:
Great to see this! :rocket:

backend/src/controller/map.rs
kitzbergerg: @Bushuo Would you mind having a look? I personally feel like it is easier to just return a matrix/vector from the backend and then generating the heatmap in the frontend using a library. I found two after a quick search:
- https://www.npmjs.com/package/react-heatmap-grid (seems simple to use)
- https://www.npmjs.com/package/heatmapjs (looks quite nice but no idea if compatible with react)

I would then build the API according to your wishes (I don't care if its an array, matrix, contains objects with keys, etc.).
You can use the endpoint as it currently is. It will simply return dummy data (0.5 in one corner; 0 everywhere else). Documentation can be found in swagger.

Bushuo: @markus2330 Yes, sending via parameters is good. I did it like this for the timeline use case.

@kitzbergerg I think an matrix is better than a image. I think, with an array the frontend can more easily adjust the scale and crop the invisible parts related to the viewport size than with an image.
Thank you for searching the libraries, but i don't know if they will help us.
I thought of directly overlaying the heat-map in a separate layer inside of konva. Does not seem too difficult to implement from scratch.

markus2330: We decided today that we want to stay with a picture in the format PNG (more efficient than a textual matrix). It should be the same size as the base layer image, so that the scaling of the base layer can be reused.

kitzbergerg: @Bushuo I now changed it to return an image. The image itself is really small as it is not clear in the backend how large it should be. So I would suggest leaving it like this in the backend and then scaling in the frontend. Otherwise different screen sizes, resolutions etc might lead to problems (I have no idea how image scaling would work in the backend). Is this ok for you or do you have a solution about how to set the size in the backend?

@markus2330 How should the size of the map be set? I think we would need something like a polygon selector when creating the map so it actually works. For now I just create dummy values in the frontend that will be sent when creating a map.
Additionally how does when merging and deploying? As soon as this is merged all previously created maps will be invalid as the geometry has a non null constraint?

kitzbergerg:
@Bushuo There is an error in storybook. It seems to also scan the generated definitions file which contains an example of the geometry in json. Would you mind fixing that? Probably the best solution would be to exclude the definitions.ts from storybook somehow, if that is not possible I don't really know how to handle that.

backend/src/controller/map.rs
markus2330: > I think we would need something like a polygon selector when creating the map so it actually works.

Yes, this is the plan, unfortunately the polygon selector is not yet implemented.

> Additionally how does when merging and deploying?

I don't understand the question.

> As soon as this is merged all previously created maps will be invalid as the geometry has a non null constraint?

This is currently no problem, we can reset the databases of dev and www.

markus2330: If there isn't a polygon yet, we probably want/need a default polygon. I suggest we use a square 100.000x100.000 Pixel, i.e. 1000x1000m, i.e. 1ha. This is also the maximum size allowed. One pixel of the heat map should have 10x10 Pixel of the Konva map, i.e. 10x10cm rastering. Please also add this to docu.

kitzbergerg: I thought the coordinates where in meters? 
So for example a planting has x and y coordinates. They have to be whole number as they are defined as i32, so if the matrix should be in 10x10cm should the plantings coordinates be in cm?

markus2330: 1px of kanvas = 1cm, unfortunately only written in doc/guidelines/ui.md

Bushuo:
> @Bushuo There is an error in storybook. It seems to also scan the generated definitions file which contains an example of the geometry in json. Would you mind fixing that? Probably the best solution would be to exclude the definitions.ts from storybook somehow, if that is not possible I don't really know how to handle that.

@kitzbergerg Hi, I looked at it. 
I don't want to fix it, as it is quite complicated. 
It can be avoided though, by putting the json code inside of backticks like `{"rings":[[{"x":0.0,"y":0.0},{"x":5.0,"y":0.0},{"x":5.0,"y":5.0},{"x":0.0,"y":5.0},{"x":0.0,"y":0.0}]],"srid":4326}`
This will tell mdx to treat it as a code block and thus ignore any reserved characters.

kitzbergerg:
@markus2330 This is now finished I would say. It is a bit difficult to test as many things are not yet in place, but I will provide some scripts that can be used.

@markus2330 @Bushuo Here are a few issues/implementation details that appeared related to this:
- I believe the (x,y) coordinates of plantings are set wrong in the frontend. Coordinates in PostGIS are (0,0) at the bottom left, so this needs to be fixed.
- Plant relations cannot yet be inserted into the database with the scraper (unless I missed something).
- The polygon for the map still has to be created in the frontend. Currently it is just a hardcoded dummy implementation.
- The generated map needs to be moved and scaled in the frontend.


# Steps
Here are the steps I used to test the heatmap:
1. Create the database.
2. Insert the plants using the scraper (`cd scraper && npm run insert`)
3. Start the backend.
4. Create a map (can be done in the frontend as well).
```bash
curl --location 'http://localhost:8080/api/maps' \
--header 'Authorization: Bearer <token>' \
--header 'Content-Type: application/json' \
--data '{
    "name": "Test1",
    "creation_date": "2023-07-06",
    "is_inactive": false,
    "zoom_factor": 100,
    "honors": 0,
    "visits": 0,
    "harvested": 0,
    "privacy": "public",
    "description": "",
    "geometry": {
        "rings": [
            [
                {
                    "x": 0.0,
                    "y": 0.0
                },
                {
                    "x": 100.0,
                    "y": 0.0
                },
                {
                    "x": 100.0,
                    "y": 100.0
                },
                {
                    "x": 50.0,
                    "y": 100.0
                },
                {
                    "x": 50.0,
                    "y": 50.0
                },
                {
                    "x": 0.0,
                    "y": 50.0
                },
                {
                    "x": 0.0,
                    "y": 0.0
                }
            ]
        ],
        "srid": 4326
    }
}'
```
5. Insert some dummy relations and plantings
```SQL
INSERT INTO relations
VALUES (1, 2, 'companion'),
       (1, 3, 'neutral'),
       (1, 4, 'antagonist');

INSERT INTO plantings
VALUES ('00000000-0000-0000-0000-000000000000', 2, 1, 10, 10, 0, 0, 0, 0, 0),
       ('00000000-0000-0000-0000-000000000001', 2, 2, 11, 10, 0, 0, 0, 0, 0),
       ('00000000-0000-0000-0000-000000000002', 2, 2, 23, 10, 0, 0, 0, 0, 0),
       ('00000000-0000-0000-0000-000000000003', 2, 2, 40, 10, 0, 0, 0, 0, 0),
       ('00000000-0000-0000-0000-000000000004', 2, 2, 10, 20, 0, 0, 0, 0, 0),
       ('00000000-0000-0000-0000-000000000005', 2, 3, 10, 20, 0, 0, 0, 0, 0);
```
6. Execute the request (Probably better to use swagger or Postman otherwise you can't see the image).
```bash
curl --location 'http://localhost:8080/api/maps/1/layers/plants/heatmap?plant_id=1&layer_id=2' \
--header 'Authorization: Bearer <token>'
```

Bushuo:
@kitzbergerg 
> I believe the (x,y) coordinates of plantings are set wrong in the frontend. Coordinates in PostGIS are (0,0) at the bottom left, so this needs to be fixed.

Yes, placement is indeed currently wrong. We changed the plant from a square to a circle which broke it. 
It was intended though, that (x,y) is the plantings center point.

Should this be the bottom left of the planting?
I don't understand, please explain.

kitzbergerg:
> Yes, placement is indeed currently wrong. We changed the plant from a square to a circle which broke it. It was intended though, that (x,y) is the plantings center point.
> 
> Should this be the bottom left of the planting? I don't understand, please explain.

The (x,y) centered in the middle of the image is fine.
The placement on the map in general is wrong.
When placing a planting in the bottom left corner the (x,y) coordinates are currently (x=0,y=800). It should however be (x=0,y=0), so that the bottom left corner has (x=0,y=0) values.

Bushuo:
> The (x,y) centered in the middle of the image is fine.
The placement on the map in general is wrong.
When placing a planting in the bottom left corner the (x,y) coordinates are currently (x=0,y=800). It should however be (x=0,y=0), so that the bottom left corner has (x=0,y=0) values.

Ok, I dont know if we can really change that in the frontend. Konvas coordinate system goes from left to right -x to x and top to bottom -y and y.
You can view it in this example of the konva docs https://konvajs.org/docs/posts/Position_vs_Offset.html

kitzbergerg:
> Ok, I dont know if we can really change that in the frontend. Konvas coordinate system goes from left to right -x to x and top to bottom -y to y. You can view it in this example of the konva docs https://konvajs.org/docs/posts/Position_vs_Offset.html Drag the shapes around.

You probably can do something like `y = Math.abs(y-y_max)` when creating/displaying plantings. That way it is basically flipped right?
I think it would be better to do that in the services of the frontend, that way the backend has one consistent view of its data, but I'm willing to discuss that (it might be possible to save data in PostGIS using the 'wrong' coordinate system, but that might lead to errors in the long term).

Although that is currently just for plantings, but when stuff like shade is implemented, the polygon needs to be the same as defined in PostGIS anyways, so that's something to think about as well.

kitzbergerg:
> You probably can do something like `y = Math.abs(y-y_max)` when creating/displaying plantings. That way it is basically flipped right?

@Bushuo Another way might be to define the center of the screen as (0,0). It don't think there is anything preventing that in the backend as the map can be generated the same way, you just have to move it to the right place in the frontend. Everything in the polygons and stuff like that is/can be stored as float, so that would allow for negative number.
You can then simply do the following in the frontend to transform the coordinate: `y = -y` (might be annoying for polygons, but should still be fine).

markus2330:
Please let us use the same coordination system throughout the stack, as defined by Konva. Flipping it somewhere is a questionable idea.

kitzbergerg:
> Please let us use the same coordination system throughout the stack, as defined by Konva. Flipping it somewhere is a questionable idea.

Done.

> * [x]  I believe the (x,y) coordinates of plantings are set wrong in the frontend. Coordinates in PostGIS are (0,0) at the bottom left, so this needs to be fixed. (Done by changing the coordinate system in the backend to be (0,0) at top left.)
> * [ ]  Plant relations cannot yet be inserted into the database with the scraper (unless I missed something).
> * [ ]  The polygon for the map still has to be created in the frontend. Currently it is just a hardcoded dummy implementation.
> * [ ]  The generated map needs to be moved and scaled in the frontend.

So there are still these things left for it to fully work.
@markus2330 @Bushuo Please assign/create issues for those tasks and link them here (I don't know if there are already issues for that).
I think the second one will be done by @temmey right?

Bushuo:
I added this issue related to the last point in the list #592

temmey:
> I think the second one will be done by @temmey right?

Is in pr #606

markus2330:
> Here are the steps I used to test the heatmap:

Please add this test protocol to doc/tests

backend/src/controller/plant_layer.rs
```diff
@@ -1,12 +1,53 @@
...
-    config::data::AppDataInner, model::dto::RelationSearchParameters, service::plant_layer,
+    config::data::AppDataInner,
+    model::dto::{HeatMapQueryParams, RelationSearchParameters},
+    service::plant_layer,
 };
 
+/// Endpoint for generating a heatmap signaling ideal locations for planting the plant.
+///
+/// The heatmap will be an exact fit on the maps geometry.
+/// This means if the maps y_min is 100 the generated heatmap will need to be moved to start at y=100.
```
markus2330, line 18: A much more detailed description of the returned image is needed.

backend/src/model/dto.rs
```diff
@@ -215,9 +216,15 @@ pub struct MapDto {
     pub location: Option<Coordinates>,
     /// The id of the owner of the map.
     pub owner_id: Uuid,
+    /// The geometry of the map.
+    ///
+    /// E.g. `{"rings": [[{"x": 0.0,"y": 0.0},{"x": 1000.0,"y": 0.0},{"x": 1000.0,"y": 1000.0},{"x": 0.0,"y": 1000.0},{"x": 0.0,"y": 0.0}]],"srid": 4326}`
```
markus2330, line 221: Please describe the geometry. Are there any required properties of that geometry? (E.g. closed)

Also describe how it looks serialized (maybe link to where it is described), e.g. what is "rings"?

Probably the description better should be in the endpoint to not duplicate the information.

backend/src/model/entity/plant_layer.rs
```diff
@@ -12,10 +17,91 @@ use crate::{
...
+    y: i32,
+}
+
+/// Generates a heatmap signaling ideal locations for planting the plant.
+///
+/// # Errors
+/// * If the SQL query failed.
+#[allow(
+    clippy::cast_sign_loss,
+    clippy::indexing_slicing,
```
markus2330, line 60: Please also add a comment here why these two are okay?

backend/src/model/entity/plant_layer.rs
```diff
@@ -12,10 +17,91 @@ use crate::{
...
+    debug!("{}", debug_query::<Pg, _>(&bounding_box_query));
+    let bounding_box = bounding_box_query.get_result::<BoundingBox>(conn).await?;
+
+    // Fetch the heatmap
+    let query = diesel::sql_query("SELECT * FROM calculate_score($1, $2, $3, $4, $5, $6, $7, $8)")
+        .bind::<Integer, _>(map_id)
+        .bind::<Integer, _>(layer_id)
+        .bind::<Integer, _>(plant_id)
+        .bind::<Float, _>(GRANULARITY)
+        .bind::<Float, _>(bounding_box.x_min)
```
markus2330, line 81: Why is the bounding box float? Better to have Integer everywhere.

backend/src/test/plant_layer_heatmap.rs
```diff
@@ -0,0 +1,205 @@
...
+    assert_eq!(resp.status(), StatusCode::OK);
+    assert_eq!(
+        resp.headers().get(header::CONTENT_TYPE),
+        Some(&header::HeaderValue::from_static("image/png"))
+    );
+    let result = test::read_body(resp).await;
+    let result = &result.bytes().collect::<Result<Vec<_>, _>>().unwrap();
+    let image = load_from_memory_with_format(result.as_slice(), image::ImageFormat::Png).unwrap();
+    let image = image.as_rgb8().unwrap();
+    assert_eq!((1, 10), image.dimensions()); // smaller by factor of 10 because of granularity
```
markus2330, line 109: Use granularity constant.

doc/tests/protocols/MTP_2023-07-8_Heatmap_V1.md
```diff
@@ -0,0 +1,104 @@
+# Test Protocol Heatmap
```
markus2330, line 1: Link test protocol from use case.

frontend/src/features/maps/routes/MapCreateForm.tsx
```diff
@@ -127,6 +127,20 @@ export default function MapCreateForm() {
       privacy: mapInput.privacy,
       description: mapInput.description,
       location: !Number.isNaN(mapInput.location.latitude) ? mapInput.location : undefined,
+      // TODO: implement selector
```
markus2330, line 130: Please describe better what you mean here. Do you mean #494, then please link to the issue.

backend/src/model/entity/plant_layer.rs
```diff
@@ -12,10 +17,91 @@ use crate::{
     schema::relations,
 };
 
+/// The resolution of the generated heatmap in cm.
+const GRANULARITY: f32 = 10.0;
```
markus2330, line 21: Great to have it but it is not used everywhere (not in tests). Why float? Better we use integer everywhere.

markus2330 (changes requested):
This is certainly a big step forwards, great to have the endpoint. I would prefer integer used more. Documentation etc. is a bit weak. Thanks for adding the test protocol, it now worked for me but I couldn't reproduce it with self-planted plantings. Probably a bigger map is needed for that (it should be 1ha per default anyways).

About the (not previously discussed) implementation in plpgsql (we actually said it should be in Rust or directly in a SQL query using PostGIS) I am not sure. The main concern is the performance: please check it as part of your thesis.

markus2330:
Usually the left-over issues should be created by the implementer of a PR.
I created them for this PR:

> * [x]  I believe the (x,y) coordinates of plantings are set wrong in the frontend. Coordinates in PostGIS are (0,0) at the bottom left, so this needs to be fixed. (Done by changing the coordinate system in the backend to be (0,0) at top left.)

#635

> * [ ]  Plant relations cannot yet be inserted into the database with the scraper (unless I missed something).

#636

> * [ ]  The polygon for the map still has to be created in the frontend. Currently it is just a hardcoded dummy implementation.

#349 

> * [ ]  The generated map needs to be moved and scaled in the frontend.

#637

frontend/src/features/maps/routes/MapCreateForm.tsx
```diff
@@ -127,6 +127,20 @@ export default function MapCreateForm() {
...
+      geometry: {
+        rings: [
```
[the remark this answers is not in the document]
kitzbergerg: I actually don't know. I didn't find documentation in PostGIS (although I haven't looked very much). All I did was build a `Polygon<Point>` struct in Rust and deserialize that to see what it would like in JSON. This was the result and it works.

frontend/src/features/maps/routes/MapCreateForm.tsx
kitzbergerg: I think this refers to #349 (#494 is a duplciate of that?).
Simply said when a map is created the user needs to specify the boundaries of the map. This boundary is then used by the heatmap (and maybe other layers like the base layer?).

frontend/src/features/maps/routes/MapCreateForm.tsx
```diff
@@ -127,6 +127,20 @@ export default function MapCreateForm() {
...
+          [
+            { x: 100.0, y: 100.0 },
+            { x: 300.0, y: 700.0 },
+            { x: 700.0, y: 700.0 },
+            { x: 1000.0, y: 1000.0 },
+            { x: 100.0, y: 1000.0 },
+            { x: 100.0, y: 100.0 },
```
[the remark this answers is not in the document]
kitzbergerg: I think this should now be 1ha.
Although until the frontend actually displays the heatmap it doesn't really matter.

backend/migrations/2023-07-03-165000_heatmap/up.sql
```diff
@@ -0,0 +1,144 @@
...
+        SELECT
+            ST_Envelope(geometry) AS bbox
+        FROM maps
+        WHERE id = map_id
+    ) AS subquery;
+END;
+$$ LANGUAGE plpgsql;
+
+
+-- Returns scores from 0-1 for each pixel of the map. It should never return values greater than 1.
```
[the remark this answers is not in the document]
kitzbergerg: Leftover. It is only between 0 and 1.

backend/migrations/2023-07-03-165000_heatmap/up.sql
```diff
@@ -0,0 +1,144 @@
...
+-- Values where the plant should not be placed are close to 0.
+-- Values where the plant should be placed are close to 1.
+-- The score at position x=0,y=0 is at the position x=x_min,y=y_min.
```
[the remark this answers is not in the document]
kitzbergerg: I hope this explains it better.

backend/src/model/dto.rs
kitzbergerg: > Please describe the geometry. Are there any required properties of that geometry? (E.g. closed)

As described above I actually don't know why it looks like this.
I think it has to be closed, so the last coordinate has to be the same as the first one, but I'm not sure.

> Also describe how it looks serialized (maybe link to where it is described), e.g. what is "rings"?

I don't know exactly what this means. The serialized struct is `Polygon<Point>`, but I don't think that's what you mean. The fields in Rust look pretty much the same as the JSON.

> Probably the description better should be in the endpoint to not duplicate the information.

I think it fits better here. The DTOs get displayed in Swagger as well, and that way it fits to the struct.

backend/src/model/entity/plant_layer.rs
kitzbergerg: Float so it is possible to have values between 0 and 1. It is probably not necessary, but doesn't hurt to have.

backend/src/model/entity/plant_layer.rs
```diff
@@ -12,10 +17,91 @@ use crate::{
...
+        heatmap[y as usize][x as usize] = score;
+    }
+    Ok(heatmap)
+}
+
 /// Get all relations of a certain plant.
 ///
 /// # Errors
-/// * Unknown, diesel doesn't say why it might error.
+/// * If the SQL query failed.
```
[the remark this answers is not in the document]
kitzbergerg: I'm not sure, but it might actually return a 404 if there are no rows returned. I don't know exactly how diesel handles that, but in some instances it returns 404 on no rows found.

doc/tests/protocols/MTP_2023-07-8_Heatmap_V1.md
kitzbergerg: I don't know what this is supposed to mean. There is no test protocol for the use case in plant layer.
Should I link to the usecase? Something like `../usecases/current/plants_layer`?

markus2330: In the use case (plants layer), there should be a link to this test protocol [doc/tests/protocols/MTP_2023-07-8_Heatmap_V1.md](https://github.com/ElektraInitiative/PermaplanT/pull/566/files/6ca9089da16b8961a3b6d5b1bc43be6fd9944446#diff-3bb78915248327a5a3bd26d3a2a3a998e6e1085c948810391bc28d0c8fe27ba3).

frontend/src/features/maps/routes/MapCreateForm.tsx
markus2330: I made a distinction between these two issues (one for backend, other for frontend).

> Simply said when a map is created the user needs to specify the boundaries of the map. This boundary is then used by the heatmap (and maybe other layers like the base layer?).

Please read the use cases, the map needs to be created without the boundaries. So it must be possible to set a boundary afterwards.

frontend/src/features/maps/routes/MapCreateForm.tsx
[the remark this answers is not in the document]
markus2330: Of course it matters, we want to implement the use cases rightly.

frontend/src/features/maps/routes/MapCreateForm.tsx
kitzbergerg: Should the map boundary be part of the base layer or part of the map? As far as I understand the following means it is part of map: `these boarders are stored in the map and not subjective to "alternatives"` (for the base_layer use case).
If so that would mean I simply need to set the field to nullable and add an option to update the boundary?

markus2330:
jenkins build please

kitzbergerg:
@markus2330 I updated it according to our meeting. I will try to finish the shade layer tomorrow, so I have enough time to actually execute the tests (I know the tests will take a while to get right, so if the shade layer which should be part of the test isn't implemented soon I won't have enough time).
As discussed I will implement/do what is now written down here, but nothing else.

badnames (approved):
Very well done.

@markus2330 Just one question: 
How often is this heatmap supposed to be updated?
As it is now, it seems to be quite resource intensive to generate.

markus2330:
> How often is this heatmap supposed to be updated?

It always gets generated freshly, as it is different for every plant and after basically every change in the map.

> As it is now, it seems to be quite resource intensive to generate.

Is this just your general feeling or do have specific concerns?

frontend/src/features/maps/routes/MapCreateForm.tsx
markus2330: Yes, the map boundary must be part of the map.

backend/migrations/2023-07-03-165000_heatmap/up.sql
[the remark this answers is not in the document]
markus2330: It is better but it could be improved: please say how exactly it needs to be moved/scaled with small pseudo code.

I created a left-over so that we can merge this: #649

PR #595, opened 2023-07-09
Add cronjob for maps deletion

### LINKED ISSUE #90: persistent undo when deleting a map/plant etc. (cron jobs)
## Goal

Functionality to revert deletion of an element within a period of time. The first version of this feature considers an element to be either a map or a plant for the time being. However, the research done here should be kept general and can be applied to other database entities in the future.

Example sequence:
1. User deletes an element on 21.03.2023 at 12:00
2. User can revert the deletion only until 21.04.2023 at 12:00

## Suggestions from Markus

>1. requires to always keep previous entries, e.g. only mark as removed+time;
>2. with partitioning to keep for at least a month

## Outcomes

Since there is no built-in job handling feature(like cron jobs in Unix systems) in PSQL, we cannot just mark entries to be removed and expect them to fade away. So table partitioning sounds more reasonable in this case from first sight. If we're going to proceed with the partitioning based on the `deleted_at` column, we can run the partitioning process either manually every day, which is 100% no-go for us, or assign it to a PSQL trigger. Unfortunately, there are only on- INSERT/UPDATE/DELETE and no time-based triggers in PSQL. This means we could only run the partitioning e.g. on the insertion of new entries which can occur within a day/week/month or even a year. This means that some entries to be removed could stay in the database for a longer period of time than they should.

My suggestions are:
1. check if we can define cron jobs on the backend level to run the deletion query
     - ```DELETE FROM plants WHERE deleted_at IS NOT NULL and deleted_at < NOW() - INTERVAL '1 month'```
2. or to combine the methods described above
     - we mark entries to be removed from a table by setting the current timestamp in the `deleted_at` column
     - define a trigger on every table change to run the deletion query(not the partitioning)
     - but the problem described above remains
3. other options?

kitzbergerg:
Introduces scheduled tasks in the backend.
Resolves #90.

## Basics

## Checklist

      (not in the PR description)
      (exceptions are documented)

## Review

- [ ] I've tested the code


4ydan (commented):
Lgtm, maybe add somewhere to the docs (doc/database/cronjobs.md or somewhere where it fits idk), that this functionality exists.

markus2330 (changes requested):
Thank you, finally it is here! Please add some documentation that this will be done and also how further cronjobs would be added.

@ThorbenStaufer can you please check if this actually delete the complete maps?

ThorbenStaufer (changes requested):
Job removes general map data, but it doesn't handle layers and placed objects and configurations on the map (plantings, base layer config, etc.). We don't have automatic delete cascading on the map entity (to my knowledge).

ThorbenStaufer (approved):
Works perfectly 👍

markus2330 (approved):
Great job, much better!

markus2330:
Needs rebase though.

markus2330:
please rebase, Changelog collided with heatmap.

markus2330:
jenkins build please

Issue #611, opened 2023-07-11
data-testid: improve naming scheme


4ydan:
There are some missing locators.
See doc/guidelines/frontend-locators.md for more information about locators.

### Tasks
- [x] Update locator guidelines to reflect naming schema (component + name)

#### (Paginated) Selectboxes:
- [x] Seed creation page both select boxes

#### Navbar:
- [x] Language selector

#### Map create page
- [x] Set location button
- [x] The privacy select box

#### Map management page
- [x] It would be nice if each map could be accessed with a locator instead of with text right now

markus2330:
Thanks for raising the issue!

Can you maybe create a CI check that verifies all elements have ids?

4ydan:
I am not sure if I should wait with everything being refactored first before I write the tests. Maybe I should also do that myself?
We already had #706 and now I have issues accessing the radio buttons in my tests.
![image](https://github.com/ElektraInitiative/PermaplanT/assets/53004139/f0195f93-d3b5-46e1-b261-25144d1167e1)
Also refactoring tests later when locators are changed is also something that has to be done.

markus2330:
You can also refactor while writing tests. But to not have all the work load on you, I think we should finish #653 and advertise it tomorrow in the meeting.

markus2330:
@4ydan is here still something to be done?

4ydan:
There is surely some missing locators, I will update the top post.

markus2330:
Ok. Isn't it better to simply extend locators as needed for E2E directly within the PRs implementing E2E tests?

4ydan:
> Isn't it better to simply extend locators as needed for E2E directly within the PRs implementing E2E tests?

We should probably decide on whos job it is to write these locators, maybe pass this responsibility to the e2e test writer and mention it in the guidelines?

markus2330:
Yes, I am okay with 

- everything that helps you and
- what we can
  -  clearly ask for using guidelines and
  - check in reviews.

markus2330:
@badnames if possible, please help @4ydan

markus2330:
Is this done?

4ydan:
No, the top post is still reflecting the missing locators.
Language selector is #904

markus2330:
@4ydan can you maybe tell @badnames which locator you want where?

We also should update the frontend locator guidelines so that it is less prone to non-uniqueness, maybe as suggested by @tschawax.

4ydan:
[...]

PR #574, opened 2023-07-04
Version number displayed on the navbar, moved github link to the footer

4ydan:
#510 

## Basics

- [x] I added a line to [/doc/CHANGELOG.md](/doc/CHANGELOG.md)
- [x] The PR is rebased with current master.
- [x] Details of what you changed are in commit messages.
- [ ] References to issues, e.g. `close #X`, are in the commit messages.
- [x] The buildserver is happy.

## Checklist

- [ ] I have installed and I am using [pre-commit hooks](../doc/contrib/README.md#Hooks)
- [ ] I fully described what my PR does in the documentation
      (not in the PR description)
- [ ] I fixed all affected documentation
- [ ] I fixed all affected decisions
- [ ] I added unit tests for my code
- [ ] I added code comments, logging, and assertions as appropriate
- [ ] I translated all strings visible to the user
- [ ] I mentioned [every code or binary](/.reuse/dep5) not directly written or done by me in [reuse syntax](https://reuse.software/)
- [ ] I created left-over issues for things that are still to be done
- [ ] Code is conforming to [our Architecture](/doc/architecture)
- [ ] Code is conforming to [our Guidelines](/doc/guidelines)
      (exceptions are documented)
- [ ] Code is consistent to [our Design Decisions](/doc/decisions)

## Review

- [ ] I've tested the code
- [ ] I've read through the whole code
- [ ] Examples are well chosen and understandable

PR #613, opened 2023-07-11
Login logout button ids

4ydan:
This adds id's to the login and logout button

## Basics

## Checklist

      (not in the PR description)
      (exceptions are documented)

## Review

- [ ] I've tested the code

### DIFF
--- a/doc/guidelines/frontend.md
+++ b/doc/guidelines/frontend.md
@@ -3,3 +3,4 @@
 - We use minimal number of packages: need to be noted in solutions or in a decision
 - Be thoughtful of memory consumption and runtime
 - Error Handling https://tkdodo.eu/blog/breaking-react-querys-api-on-purpose#a-bad-api
+- All user interactable elements should have a unique id so they can be referenced by automated tests.
--- a/frontend/src/features/auth/components/LoginButton.tsx
+++ b/frontend/src/features/auth/components/LoginButton.tsx
@@ -18,7 +18,11 @@ export const LoginButton = () => {
   if (auth.isAuthenticated) {
     return (
       <div>
-        <SimpleButton className="pl-6 pr-0" onClick={() => void auth.removeUser()}>
+        <SimpleButton
+          id="logout-button"
+          className="pl-6 pr-0"
+          onClick={() => void auth.removeUser()}
+        >
           {t('auth:log_out')}
           <ConditionalLoadingSpinner show={auth.isLoading} />
         </SimpleButton>
@@ -27,7 +31,11 @@ export const LoginButton = () => {
   }
 
   return (
-    <SimpleButton className="pl-6 pr-0" onClick={() => void auth.signinRedirect()}>
+    <SimpleButton
+      id="login-button"
+      className="pl-6 pr-0"
+      onClick={() => void auth.signinRedirect()}
+    >
       {t('auth:log_in')}
       <ConditionalLoadingSpinner show={auth.isLoading} />
     </SimpleButton>


markus2330:
If you want such little things merged fast (so that test succeeds), please add "please merge" label.

markus2330:
Please also write general guidelines in doc/guidelines/frontend.md that we want IDs everywhere.

4ydan:
> What is your opinion about querying the screen by text?
> 
> Most of modern testing frameworks support this, and it resembles more of what a user would do.
> 
I'm new to frontend testing, I haven't heard of that yet. In our specific case, we want to test a specific flow. How could we achieve that with querying text. We need also to give inputs to fields.


> Also I am not a fan of `id` where they are not necessary. Two same ids in different components could cause a bug.
> 

Makes sense. But most of the stuff like a "login-button" shouldn't exist more than once anyways.

> 
> 
> I would rather use `<div data-testid="custom-element" />`
> 
> https://testing-library.com/docs/queries/bytestid/

I am not very experienced with selenium, if it works, it's fine for me, will try it out tomorrow.
As long as I have something that I can interact with, I'm fine with every suggestion.

This what a selenium friendly website could looks like
https://www.saucedemo.com/

Bushuo:
> Oh you mean finding the login button by text? I dont think it's better than grabbing it by id.
Yeah, in this case text is worse then ids.

There is also querying by role etc, I would use ids only as a last resort, which in this case it is not.
https://autify.com/blog/why-id-should-not-be-used/

> I am not very experienced with selenium, if it works, it's fine for me, will try it out tomorrow.
I'm fine with every suggestion that just works.

Is the testing framework still up for discussion?
A really good and popular framework is playwright and you said you don't have much experience in selenium.
Maybe learning playwright is an option?
https://playwright.dev/

4ydan:
> There is also querying by role etc, I would use ids only as a last resort, which in this case it is not. https://autify.com/blog/why-id-should-not-be-used/

I agree, it's nice searching by text, because you don't need to perform additional assertions on the text afterwards.

> Is the testing framework still up for discussion? A really good and popular framework is playwright and you said you don't have much experience in selenium. Maybe learning playwright is an option? https://playwright.dev/

I mean sure, there is plenty of tools out there that basically do the same(cypress, puppeteer, etc.). What is the real benefit of playwright over selenium in our specific case. Selenium is also really good and popular. I can do it in playwright, sure. Since the performance tests are also in playwright it's maybe better. I will use playwright-python with pytest-playwright and pytest-bdd then.

markus2330:
> I mean sure, there is plenty of tools out there that basically do the same(cypress, puppeteer, etc.). What is the real benefit of playwright over selenium in our specific case. Selenium is also really good and popular. I can do it in playwright, sure. Since the performance tests are also in playwright it's maybe better. I will use playwright-python with pytest-playwright and pytest-bdd then.

Please no switching around just someone said something. Please write a proper decision and let us discuss there.

4ydan:
For me both are the same, I don't mind switching, when others benefit from it, maybe we can continue further discussion inside the issue. #611

markus2330:
> we can continue further discussion inside the issue

No, you create a decision as described in doc/decisions of this repository. We discuss in the PR adding the text of the decision. Everything someone said should be reflected in the decision (regardless of what we finally do). Please also write about the current state, i.e., what already works in your first attempt and what your experience with either framework is.

Bushuo:
@markus2330 I will look into it, we can surely import the json as module also in the respective components.

4ydan:
I can use xpaths or text meanwhile.

markus2330:
@Bushuo thanks for your input!

I created #626 to create a condensed summary of what developers need to be beware of how to implement the frontend e2e-friendly.

> I can use xpaths

Not so ideal, even super-tiny changes like #574 may already break them.

> or text meanwhile.

Also not so ideal, as it breaks even after fixing typos etc. Much better would be if we can use the i18n identifiers, see #626.

Bushuo:
> > or text meanwhile.
> 
> 
> 
> Also not so ideal, as it breaks even after fixing typos etc. Much better would be if we can use the i18n identifiers, see #626.

I think we should not discourage the use of text so hard.
Playwright and testing library can both match on Regex, which makes it more resilient.
I think stuff will be harder to refactor if we heavily rely on IDs.

markus2330:
I am not against matching on text they are definitely better than DOM matches. But using the identifiers we have for i18n is imho superior. They should change much less than the text but still say what the text is about.

If i18next discards the identifiers we could also do it the other way round: use in the e2e tests the files present in frontend/src/config/i18n/en/

It should be quite easy to write a program that given "/maps/edit/error_map_edit" returns  "Sorry, I couldn't update the map on my server. It is possible there already exists a map with the same name." (open src/config/i18n/en/**maps**.json and then we basically already have it).

In many cases (particular "Submit") I agree it does not matter much. It is probably pure coincidence if we change `/maps/edit/location_button` (we might decide the `_button` should be prefix or rename the `edit` as we get several edit fields) or `Select Location` first (for obvious reason: if we decide another text is better) first. But in case of longer texts like e.g. the error messages, it should be quite obvious that we **will** change many of them (once we start getting user feedback). And it is quite unpredictable how they are being changed.

Bushuo:
@markus2330 
> If i18next discards the identifiers we could also do it the other way round: use in the e2e tests the files present in frontend/src/config/i18n/en/

Yes, this is a nice solution.
With i18next's built-in capabilities, it is impossible to connect back from the translated string to the identifier that was used. 
For i18next it is just an argument to a function.
Additional code would be needed to also store the key alongside where it is used.
We could define a "key language", that translates key -> key or use a `data-testid` attribute where the key is passed in.

markus2330:
>   Yes, this is a nice solution.

:blush: 

> With i18next's built-in capabilities, it is impossible to connect back from the translated string to the identifier that was used.
> For i18next it is just an argument to a function.
> Additional code would be needed to also store the key alongside where it is used.
> We could define a "key language", that translates key -> key or use a data-testid attribute where the key is passed in.

Thx for the rationale, yes let us forget about it.

@4ydan what do you think about the idea to read that JSON files from testing code and use them for comparison.

@Bushuo One last thing: Are there maybe some i18n features we need to be aware of (that would break that approach)?

Bushuo:
> Are there maybe some i18n features we need to be aware of (that would break that approach)?

No, the loading of resources in our case is synchronous as well as the translation.

In the async case (loading from remote) we would need to wait for them. But we currently don't do that.

4ydan:
> @4ydan what do you think about the idea to read that JSON files from testing code and use them for comparison.

I am not sure what you mean. I would not like to read additional files, unless we really need to.

Bushuo:
> I am not sure what you mean. I would not like to read additional files, unless we really need to.

The idea was, to query the screen by the text that is displayed, which can be loaded through the json files.
Text interpolation would not work with this approach, because i18next does this during runtime.

We could also spin up an i18n instance in the tests, but we would need javascript for this.

Going forward, this all seems pretty overkill. 
Maybe the refactoring/maintenance burden is not that high and we really just use some simple ids (with `data-testid`).

4ydan:
I have no experience with that and would like to avoid it if possible.

markus2330:
I agree, new proposal in #639.

Issue #665, opened 2023-07-18
deployment on master fails


markus2330:
### Describe the Bug

https://build.libelektra.org/blue/organizations/jenkins/PermaPlanT-Folder%2FPermaplanT/detail/master/304/pipeline

### Steps to Reproduce the Problem

trigger master build

### Expected Result

all green :green_circle: 

### Actual Result

```
+ sudo /usr/local/bin/permaplant-deploy.sh dev /home/jenkins/workspace/maPlanT-Folder_PermaplanT_master
+ INSTANCE_PATH=/opt/permaplant/dev
+ BACKEND_BINARY=/home/jenkins/workspace/maPlanT-Folder_PermaplanT_master/backend/target/release/backend
+ SERVICE=permaplant-dev
+ [[ ! -d /home/jenkins/workspace/maPlanT-Folder_PermaplanT_master ]]
+ [[ ! -d /opt/permaplant/dev ]]
+ cd /home/jenkins/workspace/maPlanT-Folder_PermaplanT_master/backend
+ LC_ALL=C
+ diesel --version
diesel 2.0.1
+ cd /home/jenkins/workspace/maPlanT-Folder_PermaplanT_master/backend
+ LC_ALL=C
+ diesel migration run --locked-schema
Running migration 2023-07-03-165000_heatmap
Failed to run 2023-07-03-165000_heatmap with: column "geometry" of relation "maps" contains null values
script returned exit code 1
```

### PermaplanT version

master

### Operating System

Linux

### Browser

Firefox

4ydan:
I have seen this error in a PR that was probaly merged now.

4ydan:
This PR https://github.com/ElektraInitiative/PermaplanT/pull/622

markus2330:
Thx! Can you also add the link to the build log?

Btw. the PR must have some changes to migrations...

markus2330:
I agree.

My best guess it that the faulty one is: #595

We don't have yet the #134, so it slipped through (even though being successfully build).

If there are no other ideas, I would revert #595

4ydan:
I can try some commits out. Up to you.

4ydan:
I will checkout https://github.com/ElektraInitiative/PermaplanT/pull/595 and test it locally.

4ydan:
Yes its that branch.

[...]

Issue #206, opened 2023-04-02
Squash existing migrations.


badnames:
Since there was some confusion about how this table should look like and be named the migrations are very granular and it's hard to get an idea of our database schema by looking at them.

It might be a good idea to do a clean start by squashing them from `2023-01-20-024832_create_enums` to this one.

1. Make sure your local dev db is migrated to reflect the state of this branch.
2. Delete every migration but `00000000000000_diesel_initial_setup`.
3. Make a new migration.
4. Run `pg_dump --schema-only` and save the results in this new migration.
5. Remove diesel-related tables, functions, etc.
6. Review the leftover statements.

If we decide to do this it should be done as the last step before this is merged as things might change in the review process.
Note that diesel expects the ordering of columns in the db table and attributes in the entity struct to be the same!

_Originally posted by @Benjaminimal in https://github.com/ElektraInitiative/PermaplanT/pull/198#discussion_r1154563172_

markus2330:
Thank you for creating the issue! We can do this just before the release. After the release we cannot do this anymore, as then we want proper migrations (at least from milestone to milestone).

markus2330:
The task is already done in ./backend/schema_dump/after_squash_schema.sql it "only" needs to be moved to the actual migrations and the old migrations need to be removed.

This should probably be done together with  #233

temmey:
can be closed with #644, IMO we can also remove the schema_dump dir.

markus2330:
You can write `closes #206` in a git commit message, then it will be automatically closed.

Issue #134, opened 2023-03-21
test migrations in CI


markus2330:
We should run:

- `diesel migration redo`  **and** `diesel migration redo -a` within the **PR** CI job
- `diesel migration redo` within the **master** (dev) CI jobs
 
to test if migrations can successfully be redone.

4ydan:
Not execatly sure how to do this in the current state of the pipeline. There is no docker inside the PR and master stage, so there is no possibilities of running diesel inside those stages.
Do I stash the diesel executable over into those stages? Would that even work?

markus2330:
When you build the backend, it always runs the migrations. So the easiest way probably is to run these tests directly after building.

Btw. #644 would fail these migration tests (on purpose).

4ydan:
Ok, that was a misunderstanding from my side, I see what you want now.

markus2330:
I think the tests are not yet complete.

The test should also check if the generated schema stays the same after redoing the migrations.

Ideally, they would also test if no data gets removed.

(Changes in triggers etc. like done in #684 are out-of-scope as this can be easily seen in code reviews and the potential harm for the productive system is small.)

markus2330:
@filo14 great work, are all scenarios written in the top post tested now? Please check them.

filo14:
@markus2330 we check all of these, if that is what is meant: 
- do migration
- check schema
- do remigration
- check if schema changed compared to only one migration

PR #666, opened 2023-07-18
Revert "Add cronjob for maps deletion"

markus2330:
Reverts ElektraInitiative/PermaplanT#595


markus2330:
Should fix #665. Afterwards we can merge it again, as we will need to reset all databases for #206 anymore.

This PR is to not have to do the resets twice.

4ydan:
Working on finishing #134, then we can also test if it works.

4ydan (approved):
Seems to revert it, lets hope that was the issue.

Issue #689, opened 2023-07-22
Base Layer: fix and enable automatic scaling


badnames:
### Tasks

 Automatic scaling is currently disabled (by commenting out relevant code pieces) since its currently very buggy.
The goal of this issue is to fix and re-enable it. 

### Use case

Base layer

### Related Pull request

N/A

markus2330:
We don't need to do this for the user study, better to have https://github.com/ElektraInitiative/PermaplanT/milestone/4

Issue #661, opened 2023-07-18
Debounce base layer form


badnames:
### Tasks

 The text fields for the base layer need to be extended with Pauls debouncing facilities.
 
At the moment their data is only updated once the user deselects a textfield (=> reacts onBlur event)

### Use case

Base layer

### Related Pull request

https://github.com/ElektraInitiative/PermaplanT/pull/584

markus2330:
Thank you for creating this issue! It is one of the rather urgent issues. I added it to milestone etc.

markus2330:
Looking forward to this.

markus2330:
This is probably the most important left-over of base layer.

markus2330:
Debouncing should be the same as for dates (with green ticks).

markus2330:
@absurd-turtle does it make sense to fix this in #760?

absurd-turtle:
I put it in the current sprint again.

absurd-turtle:
Oh I think this is done already.

absurd-turtle:
I think we should open another ticket with more detailed specification.
I am not 100% sure what needs to be done here.
The hook is already in use instead of the onBlur debounce.
I guess there are no visual indications yet. Is this the only thing that is missing here?

markus2330:
Probably the visual indication is the only user-visible part. But the goal is to unify it that the debouncing works the same everywhere (with the same hook).

absurd-turtle:
I have created a new issue for this #778.

PR #760, opened 2023-07-31
basic baseLayerConfig validation

absurd-turtle:
This PR prevents sending faulty baseLayerConfig at page load.

This was the easiest fix I could do.
There are still some open questions/issues that need to be addressed.

1. The baseLayerImage config gets submitted without any user interaction at page load (might have to do with the debounce implementation idk atm). The consequence was that the config on the server gets overwritten with default values. I have not changed this yet but preventing a faulty config to be submitted fixed this bug for now.
2. I noticed a lot of dead code / comments in the BaseLayerRightToolbar component. This needs some clean up. 

## Basics

## Checklist

      (not in the PR description)
      (exceptions are documented)

## Review

- [ ] I've tested the code

### DIFF
--- a/frontend/src/features/map_planning/layers/base/components/BaseLayerRightToolbar.tsx
+++ b/frontend/src/features/map_planning/layers/base/components/BaseLayerRightToolbar.tsx
@@ -9,7 +9,7 @@ import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { FileStat } from 'webdav';
 
-function validatebaseLayerOptions(baseLayerOptions: Omit<BaseLayerImageDto, 'action_id'>) {
+function validateBaseLayerOptions(baseLayerOptions: Omit<BaseLayerImageDto, 'action_id'>) {
   const { id, layer_id, path, rotation, scale } = baseLayerOptions;
 
   if (!id || !layer_id || !path || !rotation || !scale) {
@@ -61,7 +61,7 @@ export const BaseLayerRightToolbar = () => {
       rotation: debouncedRotation,
       scale: debouncedScale,
     };
-    if (validatebaseLayerOptions(baseLayerOptions))
+    if (validateBaseLayerOptions(baseLayerOptions))
       executeAction(new UpdateBaseLayerAction(baseLayerOptions));
   }, [
     baseLayerState.imageId,
--- a/frontend/src/features/map_planning/layers/base/components/BaseLayerRightToolbar.tsx
+++ b/frontend/src/features/map_planning/layers/base/components/BaseLayerRightToolbar.tsx
@@ -1,4 +1,5 @@
 import { UpdateBaseLayerAction } from '../../../layers/base/actions';
+import { BaseLayerImageDto } from '@/bindings/definitions';
 import SimpleButton from '@/components/Button/SimpleButton';
 import SimpleFormInput from '@/components/Form/SimpleFormInput';
 import useMapStore from '@/features/map_planning/store/MapStore';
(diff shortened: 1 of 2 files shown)


markus2330:
>   The baseLayerImage config gets submitted without any user interaction at page load (might have to do with the debounce implementation idk atm). The consequence was that the config on the server gets overwritten with default values. I have not changed this yet but preventing a faulty config to be submitted fixed this bug for now.

As discussed today, we actually want the same debouncing technique everywhere. I reopened #661 (which fix maybe introduced this bug)

>  I noticed a lot of dead code / comments in the BaseLayerRightToolbar component. This needs some clean up.

Some of it might be because of #689. But cleaner code would be very much welcomed!

markus2330:
@4ydan One E2E test fails here. As said today it would be great that we also get a regression test if the base layer image/scale etc. is the same after reloading the page.

4ydan:
Very weird. A lot of tests fails here. Backend seems unreachable and login was not possible. Looks like a connection issue.

markus2330:
https://build.libelektra.org/blue/organizations/jenkins/PermaplanT/detail/PR-753/4/pipeline/ failed, too

Load is 152.43, 144.57, 124.65. Maybe caused by BBB, I stopped it, let us see.

markus2330:
I fixed the problem, cloud.permaplant.net didn't have enough RAM and probably constantly restarted some processes.

Load now back to 16.30 91.02 119.26

4ydan:
We could also reduce the cores used by the tests, to not always use all cores. What do you think?

markus2330:
No, the agents running Docker are different computers, so this wouldn't help. You can even try increasing to make the tests go faster.

4ydan:
Tests are done in 16s using already all physical cores, we could install a package and use all logical cores instead, but dont think its necessary at this point.

markus2330 (approved):
LGTM

Bushuo:
@4ydan Yes, the image representing the base layer is a canvas element.
I guess the only way to validate if it is there is to compare screenshots. 
I read that playwright supports this. 
Maybe the screenshot can be scoped to just the image, so that other UI elements do not influence the outcome.

4ydan:
I have already implemented some screenshot tests. I just dont think its the best, but when there is no other way, we have to I guess.

Bushuo:
Yes I dont think there is another way with just Playwright.

markus2330:
The test is not perfect but: simply compare the values displayed in the text fields would have already prevented the regression. So to simply check if the base layer image name+ratio is again shown after reloading.

Comparing pixel by pixel is probably quite fragile, even if we only compare the middle canvas. But we can try how it works.

frontend/src/features/map_planning/layers/base/components/BaseLayerRightToolbar.tsx
```diff
@@ -8,6 +9,16 @@ import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { FileStat } from 'webdav';
 
+function validateBaseLayerOptions(baseLayerOptions: Omit<BaseLayerImageDto, 'action_id'>) {
+  const { id, layer_id, path, rotation, scale } = baseLayerOptions;
+
+  if (!id || !layer_id || !path || !rotation || !scale) {
```
markus2330, line 15: Is this true if rotation is set to 0?

Bushuo: The condition is evaluated as true yes. Bcs 0 is falsy in JS

markus2330: Already fixed in master via #776 and deployed.

PR #793, opened 2023-08-05
first version of the proposal

temmey:
## Basics

## Checklist

      (not in the PR description)
      (exceptions are documented)

## Review

- [ ] I've tested the code

### DIFF
--- a/doc/decisions/database_plant_hierarchy.md
+++ b/doc/decisions/database_plant_hierarchy.md
@@ -34,6 +34,8 @@ See the [PSQL documentation](https://www.postgresql.org/docs/current/ddl-inherit
 
 > Table inheritance is typically established when the child table is created, using the INHERITS clause of the CREATE TABLE statement.
 
+> Rust Diesel isn't intended for that. To only select data from a specific table, and not include all child tables, we would need to use the `FROM ONLY` keyword, which is not implemented in Rust Diesel.
+
 So the inheritance is useful to deal with complex DDL structure on the startup, but will not help us to avoid bulk operations e.g. updating a column for every `variety` in the entire `genus`
 
 ### One table per taxonomy rank and one for concrete plants.
@@ -84,6 +86,28 @@ Cons:
 
 - Almost everything in the plants table needs to be nullable.
 - More complex insert and update logic.
+
+### One table per taxonomy rank and one for concrete plants. + View and custom insert/update/delete functionality
+
+[Example](example_migrations/one-table-per-taxonomy-view-functions)
+
+It's similar to `One table for taxonomy ranks and one for concrete plants` We are extending it with a view and custom functions to reduce insert and update complexity in the backend.
+
+Pros:
+
+- Inserting new plants is easy. We only need to implement minor backend changes.
+
+Cons:
+
+- Attribute overrides can only be done on the variety or cultivar level.
+- More complex insert and update logic.
+  When a species/variety is added or updated, the columns can't just be set.
+  First, we need to make sure all higher levels are in the table.
+  Then we need to check for each column value if there is a higher rank that already defines the same value.
+  Only if we can't find a match, the value should be written.
+  - We can offset this issue by implementing insert/update functions.
+    Since they are going to be complicated, long-term maintainability may be an issue.
+- Almost everything in the plants and parent tables needs to be nullable. (Is this a downside?)
 
 ## Decision
 
--- a/doc/decisions/example_migrations/one-table-per-taxonomy-view-functions/2023-03-09-194135_plant_relations/up.sql
+++ b/doc/decisions/example_migrations/one-table-per-taxonomy-view-functions/2023-03-09-194135_plant_relations/up.sql
@@ -1,0 +1,140 @@
+-- Create the "family" table
+CREATE TABLE family (
+    id SERIAL PRIMARY KEY,
+    name VARCHAR NOT NULL,
+    property1 TEXT,
+    CONSTRAINT family_name_key UNIQUE (name)
+);
+
+-- Create the "genus" table
+CREATE TABLE genus (
+    id SERIAL PRIMARY KEY,
+    name VARCHAR NOT NULL,
+    property1 TEXT,
+    family_id INTEGER NOT NULL REFERENCES family (id),
+    CONSTRAINT genus_name_key UNIQUE (name)
+);
+
+-- Create the "species" table
+CREATE TABLE species (
+    id SERIAL PRIMARY KEY,
+    name VARCHAR NOT NULL,
+    property1 TEXT,
+    genus_id INTEGER NOT NULL REFERENCES genus (id),
+    CONSTRAINT species_name_key UNIQUE (name)
+);
+
+-- Create the "variety" table
+CREATE TABLE variety (
+    id SERIAL PRIMARY KEY,
+    name VARCHAR NOT NULL,
+    property1 TEXT,
+    species_id INTEGER NOT NULL REFERENCES species (id),
+    CONSTRAINT variety_name_key UNIQUE (name)
+);
+
+-- Create the "cultivar" table
+CREATE TABLE cultivar (
+    id SERIAL PRIMARY KEY,
+    name VARCHAR NOT NULL,
+    property1 TEXT,
+    species_id INTEGER NOT NULL REFERENCES species (id),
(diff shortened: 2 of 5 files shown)


doc/decisions/database_plant_hierarchy.md
```diff
@@ -85,6 +87,28 @@ Cons:
...
+
+It's similar to `One table for taxonomy ranks and one for concrete plants` We are extending it with a view and custom functions to reduce insert and update complexity in the backend.
+
+Pros:
+
+- Inserting new plants is easy. We only need to implement minor backend changes.
+
+Cons:
+
+- Attribute overrides can only be done on the variety or cultivar level.
```
markus2330, line 102: This doesn't fulfill the requirements, we need attribute overrides on any level.

doc/decisions/database_plant_hierarchy.md
```diff
@@ -85,6 +87,28 @@ Cons:
...
+- More complex insert and update logic.
```
markus2330, line 103: But it would be hidden behind triggers? In general this sounds like a good idea, as we then don't need to keep Rust, maybe Python (E2E) and JavaScript code (scraper) in sync.

markus2330 (changes requested):
In general I like the idea but it looks like it doesn't fulfill the requirements?

doc/decisions/database_plant_hierarchy.md
temmey: this is an error, properties overrides should be possible on every level.

doc/decisions/database_plant_hierarchy.md
temmey: Yes, and the view would act just like our plant table right now, we only extend it with some more properties.

markus2330:
I like the idea more and more. Remaining questions:

What about backend compilation time? From rust we basically only see the view, don't we?

How is the performance?

How will we do the import? Will we have several CSVs for each hierarchy level?

temmey:
> What about backend compilation time? From rust we basically only see the view, don't we?

With Rust, we would handle it as a table. We need to adjust the `schema.rs` with the `schema.patch` file. 
See: https://deterministic.space/diesel-view-table-trick.html for an example.
The other tables would also be generated in the `schema.rs`.
We could remove them for now with the patch file, but in the long term, we may want to build functionality that requires these tables.

> How is the performance?
:grimacing: Well, every insert/update in the plants table would need to check up to 5 other tables with these functions. 
I am not experienced enough to make a prediction about how the performance would look. 
But since plants insert/update, as far as I know, is not something which would be excessively used except while importing, I don't think it will be a huge factor.

> How will we do the import? Will we have several CSVs for each hierarchy level?
Since we want to replace the insert functions, our current implementation should work. It may be slower, but see my answer above.

markus2330:
> How is the performance?

I mean for what we do at runtime: searching plants etc.

markus2330:
To bring it shorter to an end: I think there are only two options: either like you imagine or with everything in plants (keys point to other rows in the same table). Can you please write the pro/cons between these two in the decision?

temmey:
For now, I implemented it following this hierarchy.
image
This increases complexity as some plants are cultivars, others are varieties.
Cultivars can inherit from species and varieties. 
While solvable with my proposal, is this hierarchy necessary?
We currently identify cultivars only using ''. Is this correct? For example, in 'Solanum lycopersicum 'Lemon drop',' 'Lemon drop' is the cultivar.

doc/decisions/database_plant_hierarchy.md
```diff
@@ -34,6 +34,8 @@ See the [PSQL documentation](https://www.postgresql.org/docs/current/ddl-inherit
 
 > Table inheritance is typically established when the child table is created, using the INHERITS clause of the CREATE TABLE statement.
 
+> Rust Diesel isn't intended for that. To only select data from a specific table, and not include all child tables, we would need to use the `FROM ONLY` keyword, which is not implemented in Rust Diesel.
```
badnames, line 37:
```suggestion
> Rust Diesel isn't intended for that.
> To only select data from a specific table, and not include all child tables, we would need to use the `FROM ONLY` keyword, which is not implemented in Rust Diesel.
```

Sentences need to be written on separate lines according to our documentation guidelines.

doc/decisions/database_plant_hierarchy.md
```diff
@@ -85,6 +87,46 @@ Cons:
 - Almost everything in the plants table needs to be nullable.
 - More complex insert and update logic.
 
+### One table per taxonomy rank and one for concrete plants. + View and custom insert/update/delete functionality
+
+[Example](example_migrations/one-table-per-taxonomy-view-functions)
+
+It's similar to `One table for taxonomy ranks and one for concrete plants` We are extending it with a view and custom functions to reduce insert and update complexity in the backend and scraper.
```
badnames, line 94:
```suggestion
It is similar to `One table for taxonomy ranks and one for concrete plants`.
We are extending it with a view and custom functions to reduce insert and update complexity in the backend and scraper.
```

doc/decisions/database_plant_hierarchy.md
```diff
@@ -85,6 +87,46 @@ Cons:
 - Almost everything in the plants table needs to be nullable.
 - More complex insert and update logic.
 
+### One table per taxonomy rank and one for concrete plants. + View and custom insert/update/delete functionality
```
badnames, line 90: This title is a bit confusing in my opinion.

doc/decisions/database_plant_hierarchy.md
```diff
@@ -85,6 +87,46 @@ Cons:
...
+    However, this would often lead to empty fields and potential confusion, especially when the genus of a plant and its own species are linked to a different genus.
+
+  - Alternatively, a plant could have a single foreign key representing a plant in a higher-order category, from which it would inherit properties. However, this approach could lead to:
+    a) Circular references, where plants inherit from each other indefinitely.
+    b) Extremely long inheritance chains, with a single plant inheriting properties from numerous others.
+
+Pros:
+
+- Inserting new plants is easy. We only need to implement minor backend changes.
+- Properties overrides can be done on every level.
```
badnames, line 116:
```suggestion
- Property overrides can be done on every level.
```

doc/decisions/database_plant_hierarchy.md
```diff
@@ -85,6 +87,46 @@ Cons:
...
+Cons:
+
+- More complex insert and update logic.
+  When a species/variety is added or updated, the columns can't just be set.
+  First, we need to make sure all higher levels are in the table.
+  Then we need to check for each column value if there is a higher rank that already defines the same value.
+  Only if we can't find a match, the value should be written.
+  - We can offset this issue by implementing insert/update functions.
+    Since they are going to be complicated, long-term maintainability may be an issue.
+- Almost everything in the plants and parent tables needs to be nullable. (Is this a downside?)
```
badnames, line 128: Yes I think so. If everything is nullable (even if this goes against our data model) we need to perform more manual integrity checks.

doc/decisions/example_migrations/one-table-per-taxonomy-view-functions/2023-03-09-194135_plant_relations/up.sql
```diff
@@ -0,0 +1,165 @@
...
+    CONSTRAINT check_variety_or_species CHECK (
+        variety_id IS NULL OR species_id IS NULL
+    )
+);
+
+-- Create a view joining the tables together
+-- COALESCE function accepts an unlimited number of arguments.
+-- It returns the first argument that is not null.
+
+--todo
```
badnames, line 67: Please use english here, even if you just leave notes for yourself.
This would help non german speaking team members to gauge your progress more when reviewing draft PRs.

badnames (commented):
Looks good thus far.
As usual here are a few things I noticed while reading over your changes.

chr-schr (commented):
Overall I like the proposal if overriding plant properties on every rank is something that we need.
Select performance would be great if we use a materialized view. Currently we only insert our plants offline and never update them, so the materialized view would never need to be refreshed. 
Or are we expecting that to change in the near future? (Inserting/updating plants live)

As far as I understand, maintainability will suffer a bit. Adding a new plant property would then additionally involve:  
- Adding a new column to all plant rank tables (families, genera, species, varieties, cultivars)
- Updating the plants_view to include the new column
- (Maybe) updating the insert / update trigger functions to support the new column

PR #824, opened 2023-08-15
Feature/802 create nextcloud shares on map creation

absurd-turtle:
This PR extends the map creation endpoint by creating a Nextcloud directory, Nextcloud circle and sharing the directory with the circle.

Currently the request is not transactional which means that when one of the Nextcloud requests or the database request fails, the data sources are left in an inconsistent state.

Furthermore the frontend can be refactored to make use of react query mutations.
As well as user indications like a loading spinner would be very useful.

## Basics

## Checklist

      (not in the PR description)
      (exceptions are documented)

## Review

- [ ] I've tested the code

### DIFF
--- a/backend/src/service/nextcloud.rs
+++ b/backend/src/service/nextcloud.rs
@@ -1,0 +1,284 @@
+use crate::error::ServiceError;
+use actix_http::{Method, StatusCode};
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+const BASE_URL: &str = "https://cloud.permaplant.net/";
+
+#[derive(Debug, Deserialize)]
+pub struct CreateCircleResponse {
+    pub meta: Meta,
+    pub data: CircleData,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Meta {
+    pub status: String,
+    pub statuscode: i32,
+    pub message: String,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct CircleData {
+    pub id: String,
+    pub name: String,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub struct FindCircleData {
+    pub element: FindCircleDataElement,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FindCircleDataElement {
+    pub id: String,
+    pub user_id: String,
+    pub user_type: i32,
+    pub display_name: String,
+    pub instance: String,
+}
+
+fn extract_id_from_doc(doc: &str) -> String {
+    let res: CreateCircleResponse = quick_xml::de::from_str(doc).unwrap();
+    log::debug!("{:?}", res);
+    res.data.id
+}
+
+fn extract_id_from_circle_search_response(doc: &str) -> String {
+    let res: FindCircleResponse = quick_xml::de::from_str(doc).unwrap();
+    log::debug!("{:?}", res);
+    res.data.element.id.clone()
+}
+
+pub async fn setup_nextcloud_shares(
+    user_id: Uuid,
+    map_name: &str,
+    token: &str,
+    _should_share_publicly: bool,
+) -> Result<(), ServiceError> {
+    // create Nextcloud circle with the map name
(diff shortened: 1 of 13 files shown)


backend/src/service/nextcloud.rs
```diff
@@ -0,0 +1,284 @@
...
+) -> Result<(), ServiceError> {
+    // create Nextcloud circle with the map name
+    let circle_id = create_nextcloud_circle(map_name, token).await?;
+    // create a directory with the map name
+    create_directory(user_id, map_name, token).await?;
+    // create a directory for the base layer within
+    create_directory(user_id, &format!("{}/Base/", map_name), token).await?;
+    share_directory(map_name, token, circle_id).await?;
+
+    // FIXME: nextcloud returns 500 error
```
Bushuo, line 69: Here is still a little leftover because we don't know why nextcloud failed here.

markus2330:
Thx for the update Bushuo! :heart: 

What is the current status on this? Did something additionally happen? Was Nextcloud already reconfigured?

absurd-turtle:
> Thx for the update Bushuo! ❤️
> 
> What is the current status on this? Did something additionally happen? Was Nextcloud already reconfigured?

As far as I know Nextcloud is not yet configured. @4ydan can you please take car of this.

The tests also need to be fixed still and there is one failing Nextcloud requests that needs needs to be investigated.

@Bushuo and @horenso had a look at the transaction thing and refactored the code a little bit.

markus2330:
@absurd-turtle thx for the update

Ok, good that horenso could continue working here. I think it is rather low-prior for now, though, as a good solution does not seem to be realistic soon anyway.

@absurd-turtle can you please ask the Nextcloud folks about this problem and post a link here (to where you asked)? Please also ask the question about how to rename the share for other users.

And can you please create an issue about which CORS setting needs to be changed so that the /chat API works? Btw. for me chat creation does not even work with `--disable-web-security`. Is there something I need to know about User IDs? (Please write that in the CORS issue how we can test if the CORS setup works.)

absurd-turtle:
@markus2330 I have just discovered another approach we could try for the map directories:
https://github.com/nextcloud/groupfolders/

With the group folders one can set the permissions of a subdirectory so we could create one PermaplanT directory and create subdirectories there.
If these subdirectories show up in the correct directory we can scrap the old approach and use groups and groupfolders instead of circles and sharing.
The approach would require more carefulness in the backend though because as far as I understood only the admin can create groups and add members to it but maybe I am mistaken.

absurd-turtle:
@markus2330 Regarding the CORS, it's hard to tell without trial and error...
Basically the OPTIONS request has to receive 200 OK as a response.
This has to be configured for all the requests that are needed.
Some of them might require some headers to be set additionally.

markus2330:
> I have just discovered another approach we could try for the map directories

Great discovery. I think this is the way to go. We can try to configure the first share manually and see how it works.

> Regarding the CORS, it's hard to tell without trial and error...

Thx for the input, I created #850 so that we can start with investigations.

backend/src/service/nextcloud.rs
```diff
@@ -0,0 +1,284 @@
...
+        share_with: circle_id.to_owned(),
+    };
+
+    log::debug!("Sharing directory");
+    log::debug!("{:?}", share_options);
+
+    // For public maps share the map directory with PermaplanT circle
+    let url = BASE_URL.to_owned() + "ocs/v2.php/apps/files_sharing/api/v1/shares";
+    let client = reqwest::Client::new();
+    let res = client
```
Bushuo, line 114: Instead of recreating the http client each time we want to make a request, we should use a global http client across the whole app. 
In an environment under load, this could otherwise cause socket exhaustion. https://www.aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/
Even though we are not at that scale yet, it is probably better to already think about it.

markus2330: Great suggestion, thank you! I added this to backend coding guidelines b597457c7b56fb4d0833d484c7744d14dd807eb9

markus2330:
I close this for now as we are not actively working on it now. Will pick it up later.

PR #836, opened 2023-08-17
Complete seed overview

badnames:
## Basics

## Checklist

      (not in the PR description)
- [x] I fixed the introduction tour
- [x] I wrote migrations in a way that they are compatible with already present data
- [x] I added unit tests for my code
- [x] I created left-over issues for things that are still to be done
      (exceptions are documented)

## Review

- [ ] I've tested the code


doc/tests/manual/protocol.md
```diff
@@ -297,6 +297,35 @@
...
 
+## TC-018 - Edit seed
+
+- Description: Edit seed.
+- Preconditions:
+  - User is on the view seed page.
+- Test Steps:
+  1. Press Button "Edit seed"
+  2. Change an attribute of the seed.
+  3. Submit the form.
```
markus2330, line 308: Unfortunately, pressing "edit seed" is still broken on pr.permaplant.net:
`Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://api/seeds/2. (Reason: CORS request did not succeed). Status code: (null).`

doc/tests/manual/protocol.md
```diff
@@ -297,6 +297,35 @@
...
+- Test Result:
+- Notes:
+
+## TC-019 - Delete seed
+
+- Description: Edit seed.
+- Preconditions:
+  - User is on the view seed page.
+- Test Steps:
+  1. Press Button "Delete seed"
```
markus2330, line 322: Same problem but different reaction (via Toast instead of popup). Please use Toasts everywhere.

frontend/src/components/ExtendedPlantDisplay.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { ExtendedPlantsSummary } from '@/utils/ExtendedPlantsSummary';
+
+const ExtendedPlantsSummaryDisplayName = ({ plant }: { plant: PlantsSummaryDto }) => {
+  const transformedPlant = new ExtendedPlantsSummary(plant);
+
+  return (
+    <span>
+      {transformedPlant.displayName.common_name ? (
+        <>
+          {transformedPlant.displayName.common_name} (
```
markus2330, line 11: Great job to have this code at a single spot. Please also add additional name here, see doc/database/guidelines.md (Could be follow up but I would prefer if it is already done here, as it is difficult to test without being able to read the additional name).

frontend/src/components/Form/PaginatedSelectMenu.tsx
```diff
@@ -39,6 +39,8 @@ export interface PaginatedSelectMenuProps<
   control?: Control<T, unknown>;
   /** If set to true, the user has to select an option before the form can be completed (default false).*/
   required?: boolean;
+  /** Force a selected option. */
```
markus2330, line 42: Please elaborate a bit, does this mean you cannot select at all?

frontend/src/config/i18n/en/seeds.json
```diff
@@ -35,5 +41,6 @@
   "yield": "Yield",
   "price": "Price",
   "generation": "Generation",
-  "notes": "Notes"
+  "notes": "Notes",
+  "error_modal_title": "Fehler"
```
markus2330, line 45: Shouldn't be needed if we use a toast?

frontend/src/features/seeds/api/deleteSeed.ts
```diff
@@ -0,0 +1,12 @@
+import { baseApiUrl } from '@/config';
+import { createAPI } from '@/config/axios';
+
+export const deleteSeed = async (id: number) => {
+  const http = createAPI();
+
+  try {
+    await http.delete(`${baseApiUrl}/api/seeds/${id}`);
```
markus2330, line 8: baseApiUrl somehow seems to be wrongly used.

frontend/src/features/seeds/api/editSeeds.ts
```diff
@@ -0,0 +1,12 @@
+import { NewSeedDto, SeedDto } from '@/bindings/definitions';
+import { baseApiUrl } from '@/config';
+import { createAPI } from '@/config/axios';
+
+export const editSeed = async (seed: NewSeedDto, id: number) => {
+  const http = createAPI();
+  try {
+    await http.put<SeedDto>(`${baseApiUrl}/api/seeds/${id}`, seed);
```
markus2330, line 8: same here

markus2330 (changes requested):
Great job, looks very nice and is nearly complete.

- Unfortunately, it does not work on pr.permaplant.net, see individual comments.
- Please also write test reports, so that it is clear where/how you tested it. Even better would be if you start writing E2E tests, as this would have caught this problem.
- Please also request reviews from others to reduce the number of reviews I need to do.

frontend/src/components/Form/PaginatedSelectMenu.tsx
badnames: Essentially yes. It seems like you can't set a option that is selected by default without tracking the selected component externally to the react-select component.

doc/tests/manual/protocol.md
badnames: If i remember correctly this is a known issue in our testing environment. The backends CORS config is setup to work on permaplant.net only. pr.permaplant.net is a different origin, causing the browser to throw CORS errors.

This is the reason why I always use chromium with disabled web security when testing locally.

@4ydan Could we set the backends server URL using an environment variable to avoid such issues in the future?

markus2330: CORS is correctly set up for pr, dev and www to work with cloud (at least for getting/putting images, there might be problems for other situations).

The problem here obviously is that the URL is broken: `https://api/seeds/2` obviously is not what we want.

> This is the reason why I always use chromium with disabled web security when testing locally.

This is unrelated. For localhost+cloud CORS is not (and probably cannot be) set up. But this has nothing to do with wrong usage of the VITE_BASE_API_URL (which is set to / on pr/dev/www). I agree this should be better documented in doc/ci.md

frontend/src/components/Form/PaginatedSelectMenu.tsx
markus2330: Please add this to documentation.

frontend/src/config/i18n/en/seeds.json
badnames: I use this as a placeholder in case there was an error fetching a plants name.

markus2330:
Thank you, this is a great step forward. It is now usable. Further small things:

- [ ] after editing seeds, I still see the old seed
- [ ] seed details doesn't show which plant it is
- [ ] seed details doesn't translate the enum values `enough`, ...
- [ ] In seed edit form, `Select...` is still untranslated
- [ ] search function only works on additional name, not on plant name, okay if this is somehow shown to the user
- [ ] usability: as it is rather rare that you add a seed without having it, please revert the order, `nothing` should be last, `more than enough` should be preselected
- [ ] when in planting layer, shown seeds have the wrong name (without additional name)

Please add a realistic amount of seeds for your case study so that you can catch more of these problems.

If something is difficult to fix, please add them to a left-over issue.

markus2330:
@yvonnemarkl and me looked again and we came to the conclusion it is better to get rid of the buggy page that is shown when clicking on a specific seed. Instead it should directly go to edit the seed. To put this into a list of actionable items:

- [ ] when clicking on seed in overview, directly go to edit seed
- [ ] add two symbols (edit and trash) in the overview that let you edit and delete seeds
- [ ] ideally there should be an undo for some time before actually deleting it
- [ ] search should search on parts of the whole plant name as shown in overview
- [ ] sorting should be the same as in the map: the one that expire soonest should be at the top, using harvest year if no expire date (`Best by`) given

[Above points](https://github.com/ElektraInitiative/PermaplanT/pull/836#issuecomment-1693587470) still apply (except of the bugs related to this page when clicking on a seed, which I now ticked).

badnames:
> when in planting layer, shown seeds have the wrong name (without additional name)

@markus2330 If this is ok for you I would like to put this into a follow up issue to keep this PR on topic.

badnames:
> when clicking on seed in overview, directly go to edit seed
> add two symbols (edit and trash) in the overview that let you edit and delete seeds

@markus2330 Do you really want both?

markus2330:
>  @markus2330 If this is ok for you I would like to put this into a follow up issue to keep this PR on topic.
It shouldn't be too difficult to fix however.

Yes, it is okay okay. I only mentioned it immediately for code reuse: Both the shown names and the sort order should be identical in the seeds overview and in the seeds shown bottom right of the planting layer. 

>  @markus2330 Do you want both?
I'd prefer it if we only kept the buttons.
Handling click events on table rows as well as multiple buttons seems to be very finicky.

Yes, good that you ask: only buttons is fine. Look&feel is now very nice.

I found a little bug:

- [ ] When leaving the `More than enough` unchanged, I cannot create the seed.
- [ ] There also is an `x` next to the text, which shouldn't be there.
- [ ] After editing a seed, I again go to the site that shouldn't exist

badnames:
Thanks, I am aware of these bugs and will fix them still today hopefully.

absurd-turtle (approved):
I have only a few minor notes.
Code looks good to me now otherwise :)

frontend/src/features/seeds/components/CreateSeedForm.tsx
```diff
@@ -76,6 +131,22 @@ const CreateSeedForm = ({ isUploadingSeed, onCancel, onChange, onSubmit }: Creat
     };
   };
 
+  /**
+   *  Convert a plantId into an option for PaginatedSelectMenu.
+   *  @param plantId The id that will be converted.
+   */
+  const loadInitialPlant = async (plantId: number | undefined) => {
+    if (!plantId) return;
```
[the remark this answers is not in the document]
badnames: Undefined is now removed and replaced with an undefined check around the function call.

markus2330:
jenkins build please

Issue #872, opened 2023-08-28
seed use case left-overs


markus2330:
### Tasks

- [x] #876
- [x] offer undo for 5 sec before actually deleting a seed
- [x] sorting should be the same as in the map: the one that expire soonest should be at the top, using harvest year if no expire date (`Best by`) given
- [x] Menge/Qualität (e.g. enough and organic) not translated in Overview
- [x] reload of page+greeting when deleting plant (wanted: simply remove element without reload)
- [x] larger notes fields
- [x] Harvest Year needs two lines (should be on one line)
- [x] hover text for the two actions, "archive seed" and "edit seed"
- [x] example for price `2.99€` should be just `2.99`
- [x] example for yield should be good/hoch (lower-case)
- [x]  The paginated select results on creation are not same as the names on the management page.
- [x]  #878
- [x] UC done

badnames:
Seed ordering has been fixed in #882.

markus2330:
> Seed ordering has been fixed in https://github.com/ElektraInitiative/PermaplanT/pull/882.

Is this fix only in the plants layer or is this also for the seeds overview?

I added two more left-overs in the top post:

> 1) larger notes fields which also supports Markdown, see also #722

For now, can you maybe simply make the field bigger so that the user can write multiple lines of normal text in the notes?

There is also a visual problem which might be easily fixable:

> 2) Harvest Year needs two lines

![image](https://github.com/ElektraInitiative/PermaplanT/assets/7949075/6ec83f1f-5858-4e19-b79f-33cf4717e050)

![image](https://github.com/ElektraInitiative/PermaplanT/assets/7949075/d7b48fb8-51cd-457f-ba36-7df3ee69614a)

markus2330:
@badnames I added two more minor things:

- [ ] example for price `2.99€` should be just `2.99`
- [ ] example for yield should be good/hoch (lower-case)

The little things here can all be fixed in #902, to get the release done tomorrow.

4ydan:
Might add a placeholder to

4ydan:
I find it a bit confusing that a seed has a plant name. Shouldnt we maybe call it seed name?

4ydan:
Also is that correct that the names are flipped from the search and the seed management page
![image](https://github.com/ElektraInitiative/PermaplanT/assets/53004139/7d78d8c4-2130-4f44-951b-15bdb841a693)

[...]

Issue #355, opened 2023-05-14
Frontend restructuring


Bushuo:
As discussed in #352 and https://github.com/ElektraInitiative/PermaplanT/pull/352#discussion_r1192925554 the structure in our frontend code is not always clear.
This issue will clear up the following:
- [X] (WONTDO) Create a `lib` folder that holds the configurations for libraries (to follow bulletproof react - see [Frontend Architecture Pattern Decision](https://github.com/ElektraInitiative/PermaplanT/blob/master/doc/decisions/frontend_architecture_pattern.md))
- [x] Merge the `bindings` and `generated` folder. Such that there is only one `generated` folder that holds all generated code.
- [x] Rename `bindings` to `api_types` in the new folder structure
- [x] Remove `src/asset` and `src/icons`, instead have in `public` `icons` etc. subfolders
- [x] add all icons not drawn by us in .reuse/dep5

Bushuo:
@markus2330 Yes we conform to the above conventions.
Except for the `export default` one, which is controversial.
Here are two sources that explain the shortcomings 
https://basarat.gitbook.io/typescript/main-1/defaultisbad
https://stackoverflow.com/questions/46913851/why-and-when-to-use-default-export-over-named-exports-in-es6-modules 

I think we currently use a mix of both.
Should I update the code to just named exports?

markus2330:
Yes, I agree, we should prefer named exports. For now it is enough if we create a guideline for this and an issue about left-overs. We first need to concentrate on getting the main functionality done.

markus2330:
I added:

- [ ] Remove `src/asset` and `src/icons`, instead have in `public` `icons` etc. subfolders.

above.

markus2330:
There seems to be a problem with `Remove src/asset and src/icons`

In #881 there is now the warning:

```
Assets in public cannot be imported from JavaScript.
Instead of /public/icons/archive-off.svg, put the file in the src directory, and use /src/icons/archive-off.svg instead. (x3)
```

(although the icon is shown?)

@Bushuo What do you think about removing `src/assets` and have everything in `src/icons`? (And also all svgs in `public` would be moved there)?

markus2330:
I updated `- [ ] add all icons not drawn by us in .reuse/dep5` above.

badnames:
> Merge the bindings and generated folder. Such that there is only one generated folder that holds all generated code.

[...]

PR #881, opened 2023-08-29
Seed delete confirmation and undo.

badnames:
## Basics

## Checklist

      (not in the PR description)
- [x] I wrote migrations in a way that they are compatible with already present data
      (exceptions are documented)

## Review

- [ ] I've tested the code


markus2330:
jenkins build please

markus2330 (changes requested):
Great job, I love the undo button in the toast, looks&feel great. This also seems to fix the "Hello" greeting after removing a seed. The confirmation before, however, seems to be unnecessary and its text wrong (as you can press undo). Please remove this confirmation.

badnames:
The modal is now removed.

badnames:
@markus2330 there is a small caveat to restoring seeds I should mention:

Deleted seeds are immediately removed from the backend.
If a user chooses to redo the deletion the seed is resubmitted to our database as if it were a new seed (it gets a new id).

This does not seem to cause problems in the editor at the moment, because it ignores the seeds data completely.
However, if this changes in the future deleting and then restoring old seeds might also remove plants from maps without the user noticing.

badnames:
> This really is problematic. 

Yes I only noticed that seeds might have to keep their ids after implementing this feature.

> Isn't there a way to send the delete request only if undo wasn't pressed?

In theory absolutely.
However in practice this seems to be again much harder than initially expected.
The main issue is that the timing of the API-calls is very inaccurate.
Sometimes seeds were not actually deleted until several minutes after pressing the delete button.
 
Here is a short summary of the things I have tried locally:

### Delete the seed once react-toastify's toast.onChange callback indicates that the undo toast was closed.

Actual deletes would only happen more than 30 seconds after the toast has vanished.
While debugging I also noticed that the callback was perpetually executed in an infinite loop after this point.

This might have been an error on my side, but if it was I could absolutely not figure out what was wrong.
The documentation of this feature is unfortunately very sparse (https://fkhadra.github.io/react-toastify/listen-for-changes/). 

### Same as above but using react-toastify's legacy onClose callback.

This method did basically the opposite of what I was expecting. 
Seeds would be deleted the instant the toast appeared on the screen.
Once the toast was removed, no callback fired whatsoever (?).

According to various online sources I am apparently not the only one experiencing this.

### custom solution using setTimeout

I started a timer that executed in parallel to the toast.
Like in the first example, the callback for deleting the seed was called sometimes several seconds to minutes after it should have been called. 
This is probably just a sign that there is a lot of code running in the background that prevents the timers from executing (https://stackoverflow.com/questions/34691428/javascript-event-queue-and-settimeout).

badnames:
Possible solutions:

- Keep the current solution and document the mentioned caveat so that we can deal with it once it causes problems (probably not the best idea).
- Modify the backend in such a way that seeds can be resubmitted with old IDs
- Implement a more elaborate undo mechanism where seeds are only marked as deleted in the backend (such that they are not returned in fetch requests) with a cronjob that removes all marked seeds after a specified period of time.

markus2330:
@badnames Thx for the detailed investigations, good job! :heart: 

I talked with @yvonnemarkl and we will do the following:

- there shouldn't be a "delete" but call it "archive" instead
- the icon should be archive-off from https://tablericons.com/
- "archive" is a boolean in the seeds database
- the "archive" button and the "undo" toggle that flag in the seeds database
- as it is not a deletion, we don't need a cronjob and there cannot be plant referencing to deleted seeds

Currently, archived seeds should not be shown by default but a user can switch to show only archived seeds (similar to Nextcloud Deck). But better keep this in an left-over issue, as this feature is not urgent/release critical. As always: Please create the left-over issues as needed.

badnames:
@markus2330 I have used the icon "archive" (also from tabler icons).
This was done because the meaning of "archive-off" would usually be the opposite in other applications and websites.

If you still prefer the other icon, this can be trivially changed.

markus2330:
jenkins build please

markus2330:
jenkins build please

markus2330:
The CI didn't run through?

> @markus2330 I have used the icon "archive" (also from tabler icons).
> This was done because the meaning of "archive-off" would usually be the opposite in other applications and websites.

In our case you press this button (to archive the seed) if you **take out** the (physical) seeds from your (physical) box (or you don't have it in your physical box anymore for whatever reason), so the reverse symbol makes more sense.

The overview of (non-archived) seed should correspond 1:1 with the content of the box.

Here an update on the use case, please include in the PR:

```diff
diff --git a/doc/usecases/current/entry_list_seeds.md b/doc/usecases/current/entry_list_seeds.md
index 0c61f1b7..cf5ad945 100644
--- a/doc/usecases/current/entry_list_seeds.md
+++ b/doc/usecases/current/entry_list_seeds.md
@@ -5,7 +5,7 @@
 - **Scope:** Seed Management
 - **Level:** User Goal
 - **Actors:** App User
-- **Brief:** The users manage their seeds by adding, editing, and deleting them in the app.
+- **Brief:** The users manage their seeds by adding, editing, and archiving them.
 - **Assignee:** Moritz
 
 ## Scenarios
@@ -14,13 +14,13 @@
   The user has opened the seed management feature in the app.
 - **Main success scenario:**
   - The user is able to add new seeds to their list.
-  - The user is able to view a list of seeds where plant name, additional name, harvest year, amount, and origin is visible.
-  - The user is able to edit or delete seeds of the list.
+  - The user is able to view a list of seeds where the complete name, amount, quality, harvest year, and origin is visible.
+  - The user is able to edit or archive seeds of the list.
 - **Alternative scenario:**
-  The user accidentally deletes a seed from their list and uses the undo feature to correct it.
+  The user accidentally archives a seed from their list and uses the undo feature to correct it.
 - **Error scenario:**
-  There is an error when the user attempts to add a seed to their list, such as an invalid expiration date or a seed with the same name already existing.
-  In this case, the app validates while the user is typing and shows a validation error.
+  There is an error when the user attempts to add or edit a seed entry with invalid data such as an invalid price.
+  In this case, the app validates when saving the seed and shows a validation error.
 - **Postcondition:**
-  The user's seed list is updated with the added, deleted, or restored seeds.
+  The user's seed list corresponds exactly with the physical seeds in the user's seed box in reality.
```

markus2330:
@badnames please also add hover text for the two actions, "archive seed" and "edit seed" (to complete the archive functionality, added it to #872)

backend/src/model/enum/include_archived_seeds.rs
```diff
@@ -0,0 +1,19 @@
+//! [`IncludeArchivedSeeds`] enum.
+
+use serde::{Deserialize, Serialize};
+use typeshare::typeshare;
+
+/// Indicates whether archived seeds should be returned in a seed query.
+#[typeshare]
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+pub enum IncludeArchivedSeeds {
+    #[serde(rename = "not_archvied")]
```
markus2330, line 10:
```suggestion
    #[serde(rename = "not_archived")]
```

doc/changelog.md
```diff
@@ -8,8 +8,7 @@ Syntax: `- short text describing the change _(Your Name)_`
 
 ## 0.3.3 - UNRELEASED
 
-- _()_
-- _()_
+- Enable undo for seed deletion _(Moritz)_
```
markus2330, line 11: Does not fully match to the PR anymore.

doc/usecases/current/entry_list_seeds.md
```diff
@@ -1,26 +1,22 @@
-# Use Case: Entry and List of Seeds
```
markus2330, line 1: Something went wrong here.

markus2330 (changes requested):
functionality looks much better now, I didn't test it though. Please add test protocol.

markus2330 (approved):
Worked great in a quick test! :heart:

frontend/src/icons/archive.svg
markus2330, line 1: File not needed?

frontend/src/icons/archive-off.svg
markus2330, line 1: Should be in public, see #355. .reuse/dep5

markus2330 (commented):
Some follow ups...

4ydan:
What is the linked issue here again?

4ydan:
I am very confused, where are the archived seeds going, how can I see the archive?

badnames:
> I am very confused, where are the archived seeds going, how can I see the archive?

Archived seeds stay in the database but are marked using an additional archive date property.
At this point in time archived seeds are not displayed to the user.
In the future they will be displayed seperatelay, allowing the user to unarchive them if desired.

4ydan:
When deleting two seeds back to back you can not restore both, maybe this is a very rare edge case, but still.

markus2330:
>  It would have been better to just show a warning before deletion, as there is no archive for the user at this point.

If there is an undo, we don't need a confirmation. (#900)

>  if we put the notes field next to generation

No, notes field should get bigger, so it does not fit.

> Edit: It was because of older inserted seeds not having the archive attribute.

The PR should also work for older seeds inserted from previous versions. For old seeds, we can simply assume they are not archived. What exactly is wrong here or how is the behavior?

4ydan:
Can this be merged? @badnames

4ydan:
> The PR should also work for older seeds inserted from previous versions. For old seeds, we can simply assume they are not archived. What exactly is wrong here or how is the behavior?

Its just that you cant restore seeds without the archive attribute

markus2330:
jenkins build please

backend/src/model/dto.rs
```diff
@@ -101,6 +104,14 @@ pub struct NewSeedDto {
     pub notes: Option<String>,
 }
 
+#[allow(clippy::missing_docs_in_private_items)] // TODO: See #97.
```
markus2330, line 107: Please remove

```suggestion
```

backend/src/controller/seed.rs
```diff
@@ -132,3 +135,19 @@ pub async fn edit_by_id(
     let response = service::seed::edit(*id, user_info.id, edit_seed_json.0, &app_data).await?;
     Ok(HttpResponse::Accepted().json(response))
 }
+
+/// Endpoint archiving/unarchiving a [`Seed`](crate::model::entity::Seed).
+///
```
markus2330, line 140: Mention that the timestamp will be recorded. (which btw. is a very nice solution)

frontend/public/icons/archive-off.svg
markus2330, line 1: 1. Still not in .reuse/dep5`
2. It looks like you need to undo the revert: I get the warning ```Assets in public cannot be imported from JavaScript. Instead of /public/icons/archive-off.svg, put the file in the src directory, and use /src/icons/archive-off.svg instead. (x3)` #355

markus2330 (changes requested):
A few things still to be fixed. I also encountered a weird problem where two toasts said that the same seed was restored. I cannot reproduce it, however. But maybe you find a way how to make the code more robust.

markus2330:
> Its just that you cant restore seeds without the archive attribute

Cannot reproduce. For me archive&restoring also works for seeds that were created before this PR.

badnames:
> > Its just that you cant restore seeds without the archive attribute
> 
> Cannot reproduce. For me archive&restoring also works for seeds that were created before this PR.

At least in theory every existing seed should get the new archived_at column that is initially set to null.

markus2330:
@4ydan we also need test cases with more than 30 objects, as this is our page size.

frontend/src/icons/archive-off.svg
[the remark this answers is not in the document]
badnames: I could not find any licensing info on tablericons.com.
It only says "No attribution required" and "For comercial use".

badnames:
> @4ydan we also need test cases with more than 30 objects, as this is our page size.

@markus2330 Thank you for providing console output this time. 
This makes troubleshooting much quicker and easier.
  
Also: the error had nothing to do with the amount of seeds on the page. 
There was simply a missing enum value in the hook useTranslateQuantity (it says so right in the error message).

frontend/src/icons/archive-off.svg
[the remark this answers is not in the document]
markus2330: Yes, you can copy this text as-is, together with the URL from where we got it from.

PR #883, opened 2023-08-29
Translate quality and quantity in seed overview

badnames:
## Basics

## Checklist

      (not in the PR description)
      (exceptions are documented)

## Review

- [ ] I've tested the code

### DIFF
--- a/frontend/src/hooks/useTranslateQuality.ts
+++ b/frontend/src/hooks/useTranslateQuality.ts
@@ -1,0 +1,24 @@
+import { Quality } from '@/bindings/definitions';
+import { useTranslation } from 'react-i18next';
+
+/**
+ * Returns a function that translates values of the Quality enum.
+ */
+export function useTranslateQuality() {
+  const { t } = useTranslation('enums', { keyPrefix: 'Quality' });
+
+  return (key: Quality) => {
+    // We use a switch case here to avoid having to fight with
+    // the type system.
+    switch (key) {
+      case 'organic':
+        return t('Organic');
+      case 'not organic':
+        return t('NotOrganic');
+      case 'unknown':
+        return t('Unknown');
+      default:
+        throw 'Unknown enum key';
+    }
+  };
+}
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -61,7 +61,7 @@ Syntax: `- short text describing the change _(Your Name)_`
 - Add Christoph Nemeth as Teammember _(tschawax)_
 - _()_
 - _()_
-- _()_
+- Seeds: translate quality and quantity columns _(Moritz)_
 - _()_
 - _()_
 - _()_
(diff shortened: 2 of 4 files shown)


frontend/src/hooks/useTranslateQuality.ts
```diff
@@ -0,0 +1,24 @@
+import { Quality } from '@/bindings/definitions';
+import { useTranslation } from 'react-i18next';
+
+/**
+ * Returns a function that translates values of the Quality enum.
+ */
+export function useTranslateQuality() {
+  const { t } = useTranslation('enums', { keyPrefix: 'Quality' });
```
markus2330, line 8: Why implement it as a hook?

frontend/src/hooks/useTranslateQuality.ts
```diff
@@ -0,0 +1,24 @@
...
+    // the type system.
+    switch (key) {
+      case 'organic':
+        return t('Organic');
+      case 'not organic':
+        return t('NotOrganic');
+      case 'unknown':
+        return t('Unknown');
+      default:
+        throw 'Unknown enum key';
```
markus2330, line 21: Is it maybe possible to not have this so that we get a type error if we extend the enum with another option?

markus2330 (approved):
Thank you, works for me.

frontend/src/hooks/useTranslateQuality.ts
badnames: useTranslation is also a hook.
This means it can either be called from react components or other hooks, otherwhise react would complain.

frontend/src/hooks/useTranslateQuality.ts
badnames: I agree its not an elegant solution but I chose this approach due to useTranslation's implementation.

All keys that are declared in a json translation document are converted to sum types by react-i18n.
This means the validity of translation keys that are passed to the t function is only checked at compile time.

All strings that are not guaranteed (in a way that is verifiable by typescripts type system) to contain only translation keys will be rejected.

badnames: I have reimplemented this section using code that should work for all future enum members.

However, I had to turn off all compiler and linker warnings, since typescript can't automatically verify that the string given to the t function is a valid translation key.
This is also the reason why I went with the switch/case solution in the first place.

markus2330: It was only a question, not a request to change it.

I liked the previous solution better. I don't think we will often extend the enum. My question was only about if we can use the compiler to help us telling us that a change in the switch is needed.

@4ydan do the E2E tests actually tell us if JavaScript exceptions are thrown? I think the E2E tests should fail if there are unhandled exceptions.

badnames: >My question was only about if we can use the compiler to help us telling us that a change in the switch is needed.

No, I don't think so.

4ydan: >@4ydan do the E2E tests actually tell us if JavaScript exceptions are thrown? I think the E2E tests should fail if there are unhandled exceptions.

Currently I m not looking into return statuses and thrown exceptions.

markus2330: Thx, I created #896 for that.

markus2330:
Great job! :heart: :rocket:

PR #1040, opened 2023-11-05
Feature/1039 enable drawing plant field

tschawax:
## Basics

- [x] I added a line to [changelog.md](/doc/changelog.md)
- [x] References to issues, e.g. `close #X`, are in the commit messages and changelog.

## Checklist

- [ ] I added automated tests or a [manual test protocol](../doc/tests/manual/protocol.md)
- [ ] Exceptions to any guidelines are documented

## Review

- [ ] I've tested the code
- [ ] I've read through the whole documentation
- [ ] I've checked conformity to guidelines

### DIFF
--- a/frontend/src/features/map_planning/layers/plant/PlantsLayer.tsx
+++ b/frontend/src/features/map_planning/layers/plant/PlantsLayer.tsx
@@ -1,5 +1,8 @@
 import useMapStore from '../../store/MapStore';
+import { PlantForPlanting } from '../../store/MapStoreTypes';
 import { useIsReadOnlyMode } from '../../utils/ReadOnlyModeContext';
+import { SELECTION_RECTANGLE_NAME } from '../../utils/ShapesSelection';
+import { isPlacementModeActive } from '../../utils/planting-utils';
 import { CreatePlantAction, MovePlantAction, TransformPlantAction } from './actions';
 import { PlantLayerRelationsOverlay } from './components/PlantLayerRelationsOverlay';
 import { PlantingElement } from './components/PlantingElement';
@@ -17,6 +20,9 @@ import { createPortal } from 'react-dom';
 import { Layer } from 'react-konva';
 import { Html } from 'react-konva-utils';
 import * as uuid from 'uuid';
+
+// For performance reasons add limit for amount of plants inside a plant field
+const LIMIT_PLANT_FIELD_PLANTS = 50;
 
 const PLANT_WIDTHS = new Map<PlantSpread, number>([
   [PlantSpread.Narrow, 10],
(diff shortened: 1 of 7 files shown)


markus2330 (changes requested):
Thank you for the PR!

> the way you want to have it, i.e. drawing and persisting a rectangle with a single plant in it and a counter next to it, is currently not possible.

Yes, as said it would be okay if it exists only in the frontend, with commented out extra fields in the API calls, to be later added by Jannis.

> I anyway went for a (sub-optimal) working solution so you have sth to look at and possibly work with in the future.
At the moment, the only possible way to draw an area of plants is by "physically" creating all the plants in the selection rectangle. This is, for obvious reasons (performance, UI, usability), not optimal. Add to that, that we are also missing an endpoint which allows for bulk creation of plants.

Another problem is that there seems to be some limit, in bigger areas the amount of added plants seems to be restricted. (Not all the area gets filled.)

> We currently have no possibility to store the information in the database that a plant is actually representing a whole field of itself. Accordingly, we also can't know if a plant retrieved from the database is a plant field and should be rendered as such.

Actually you could also misuse some fields, e.g. the width/height currently can only be a few fixed values, all other values could be used to remember fields.

>  The database needs to be enhanced with the information that a plant can actually be a whole field of that plant.
 > Together with it, we could store the plant amount in that field.

> So in the end, what we need in the frontend, is an additional info in the PlantingDto if it's a plant field and/or the amount of plants in it.

This you could already do?

>  The amount of plants (= the count) could be shown in the top left corner of the rectangle - we could use the fact that we render all plants as a circle to our advantage cuz there will always be a small empty space in the rectangle's corners where the biggest visual part of the counter can be rendered (and the rest is slightly overlapping the circle which won't hurt at all).

Yes, rendering with individual plants is nice, Yvonne likes it.

>    A nice improvement would be to show the counter also in the left toolbar section, e.g. before the name (i.e. the heading of the section) like '20x Onion (Allium cepa)'.
>  Another nice-to-have would be to add the counter to the tooltip label when hovering over the plant field.

Actually simply having the number how many plants are selected would be already precious. Can you do this somehow in the left bottom toolbar?

>  Regarding resizing (scaling), I would, same as we do it for any plants, not set any restrictions for now and allow everything.

Resizing should increase the number of plants.

What we are still missing is that you can select more than one areas/plants via  shift/ctrl dragging.

markus2330 (approved):

frontend/src/features/map_planning/layers/plant/PlantsLayer.tsx
```diff
@@ -18,6 +21,9 @@ import { Layer } from 'react-konva';
 import { Html } from 'react-konva-utils';
 import * as uuid from 'uuid';
 
+// For performance reasons add limit for amount of plants inside a plant field
+const LIMIT_PLANT_FIELD_PLANTS = 50;
```
markus2330, line 25:
```suggestion
const LIMIT_PLANT_FIELD_PLANTS = 1000;
```

PR #1302, opened 2024-04-18
1113 adjust defaults new map defaults

andreicristian97:
## Summary

To make the default view of a new map make sense for the user, we change the scaling of the map polygon and we move the view rectangle to be centered on (0, 0)

## Basics

- [ ] The PR is rebased with current master
- [ ] I added a line to [changelog.md](/doc/changelog.md)
- [ ] Details of what I changed are in the commit messages
- [ ] References to issues, e.g. `close #X`, are in the commit messages and changelog
- [ ] The buildserver is happy

## Checklist

- [ ] I added automated tests or a [manual test protocol](../doc/tests/manual/protocol.md)
- [ ] Exceptions to any guidelines are documented

## First Time Checklist

- [ ] I am using [Tailwind CSS Linting](https://tailwindcss.com/blog/introducing-linting-for-tailwindcss-intellisense)

## Review

- [ ] I've tested the code
- [ ] I've read through the whole documentation
- [ ] I've checked conformity to guidelines
- [ ] I've checked conformity to requirements
- [ ] I've checked that the requirements are tested


andreicristian97:
Explanation related to how the View rectangle is centered at (0, 0):
- We need to figure out the coordinates of the top left corner of the map. For this, it is enough to calculate the actual width and height of the visible part of the map
- We use the window.innerWidth/Height properties as a base for the calculation, from which we need to subtract some values
- For the width, we need to subtract twice the default width of the Toolbar element (this is currently hardcoded to be 300 inside the component)
- For the height, we need to subtract the height of the Navbar (64px coming from the height property h-16 equalling 4rems), and the height of the Timeline (this is a bit harder to calculate step-by-step as it is a result of many styles - total height is currently 133px)

In the code we have those values hardcoded for the calculation - this is not ideal but I am not sure how else to approach this, maybe someone else has another solution.

markus2330:
> Explanation related to how the View rectangle is centered at (0, 0):

Please put this as code comment. It is nearly impossible for devs to find text in PRs later on.

hatchla (approved):
I would suggest having a config file of some sort (or adding it to an existing file where it fits) in the map_planning feature maybe for saving these sizes "globally". If not, leaving it like this is also good for now.

- [ ] I've tested the code
- [ ] I've read through the whole documentation
- [ ] I've checked conformity to guidelines
- [ ] I've checked conformity to requirements
- [ ] I've checked that the requirements are tested

andreicristian97:
I changed the approach to the problem. Inside the component, we use a reference to the node of the DOM tree already, so we can use it to grab the height and the width of the "BaseStage" component itself and update it inside the useEffect hook. 
It's better than having hard-coded values in the calculation; the only side effect is, if the Timeline loads after this calculation is done, the view rectangle might end up slightly off-center as its height would change when the Timeline renders. (but the difference is relatively small)
Event with the side effect in mind, I think this is better than the old calculation approach

hatchla (approved):

markus2330:
jenkins build please

markus2330:
jenkins build please

